PR #4, opened 2017-03-30
Slurm adapter

nickjer:
Adding a Slurm adapter.


nickjer:
I feel this is ready to be merged. I'd also like to test it at another Slurm center to check portability.

PR #5, opened 2017-03-30
LSF Adapter

ericfranz:


lib/ood_core/job/adapters/lsf.rb
```diff
@@ -0,0 +1,200 @@
...
+        batch = Adapters::Lsf::Batch.new(bin: c.fetch(:bin, ""))
+        Adapters::Lsf.new(batch: batch)
+      end
+    end
+
+    module Adapters
+      class Lsf < Adapter
+        attr_reader :batch
+
+        require "ood_core/job/adapters/lsf/batch"
```
nickjer: add to top of file?

lib/ood_core/job/adapters/lsf.rb
```diff
@@ -0,0 +1,200 @@
...
+          # Submit job
+          batch.submit_string(script.content, args: args, env: env)
+
+        rescue Batch::Error => e
+          raise JobAdapterError, e.message
+        end
+
+        # Retrieve job info from the resource manager
+        # @param id [#to_s] the id of the job
+        # @raise [JobAdapterError] if something goes wrong getting job info
```
nickjer: Where is this being raised?

nickjer: Same with below methods...

nickjer: It is just `#info` and `#info_all` that is missing it.

lib/ood_core/job/adapters/lsf.rb
```diff
@@ -0,0 +1,200 @@
+require "ood_core/refinements/hash_extensions"
+
+module OodCore
+  module Job
+    class Factory
+      using Refinements::HashExtensions
+
+      # Build the Lsf adapter from a configuration
+      # @param config [#to_h] the configuration for job adapter
```
nickjer: Does this not handle multiple clusters?

lib/ood_core/job/adapters/lsf.rb
nickjer: Ah, I see it is set in private method. Maybe move it up in the public method?

lib/ood_core/job/adapters/lsf.rb
ericfranz: Not yet. After it works I'm going to send an email to George with a list of questions about LSF. There is a `-m` argument for the cluster, but when I provide the name of the cluster as returned by the LSF command to get a list of available clusters the commands don't run.

lib/ood_core/job/adapters/lsf/batch.rb
```diff
@@ -0,0 +1,156 @@
...
+  end
+
+  # helper method
+  def parse_bjobs_output(response)
+    return [{}] if response =~ /No job found/ || response.nil?
+
+    lines = response.split("\n")
+    validate_bjobs_output_columns(lines.first.split)
+
+    lines.drop(1).map{ |job|
```
nickjer: Nothing wrong here, just thought it was odd you used `{}` for a multi-line block instead of `do end`.

lib/ood_core/job/adapters/lsf.rb
ericfranz: Yeah I need to refactor info_shared away.

lib/ood_core/job/adapters/lsf/batch.rb
```diff
@@ -0,0 +1,156 @@
...
+  # @param args [Array<#to_s>] arguments passed to `sbatch` command
+  # @param env [Hash{#to_s => #to_s}] environment variables set
+  # @raise [Error] if `bsub` command exited unsuccessfully
+  # @return [String] the id of the job that was created
+  def submit_string(str, args: [], env: {})
+    args = args.map(&:to_s)
+    parse_bsub_output call("bsub", *args, env: env, stdin: str.to_s)
+  end
+
+  # helper method
```
nickjer: Is this method part of the API for this object or should it be private?

lib/ood_core/job/adapters/lsf/batch.rb
ericfranz: > it was odd you used {} for a multi-line block instead of do end

I've actually seen arguments where in many cases this reads clearer than `do end`. There isn't a "hard fast" rule on "always use do end for multiple lines". I definitely prefer `{}` over `do end` if you will chain calls off of a map or reduce etc. In this case I could probably do either.

lib/ood_core/job/adapters/lsf/batch.rb
ericfranz: Its the batch object, not the adapter. I'd like to have the helpers be public. Its easier to test without mocks. It might make more sense though to move this to a batch_helper.rb class.

lib/ood_core/job/adapters/lsf.rb
ericfranz: No. If I do that, then the initial creation of the class Lsf occurs in the Lsf::Batch object.

Then I can't get away with doing this in Lsf::Batch:

```ruby
# lsf/batch.rb
class OodCore::Job::Adapters::Lsf::Batch
  # code here
end
```

I'd have to do:

```ruby
# lsf/batch.rb
module OodCore
  module Job
    module Adapters
      class Lsf < Adapter
        class Batch
          # code here
...
```

Though I could get away with this instead:

```ruby
# lsf/batch.rb
module OodCore
  module Job
    module Adapters
      class Lsf < Adapter
      end
    end
  end
end

class OodCore::Job::Adapters::Lsf::Batch
  # code here
end
```

It was easier to just move the require statement below.

ericfranz:
The separation into two objects loosely follows the design of pbs_ruby and the slurm adapter, but I have qualms about it.

To instantiate it the adapter accepts a "batch:" argument and this object's interface is likely to change in the future. Which means technically the interface of the adapter itself will change (i.e. the initializer).

Fortunately, the primary way an individual is supposed to instantiate an adapter is by using the adapter factory method instead.

nickjer:
I am fine with labeling the `#initialize` private. Especially since it isn't defined in our abstract adapter that all adapter classes inherit from.

ericfranz:
I want to do several things with this, including:

1. change the code so the tests aren't using mocks. the tests seem too coupled with the implementation and don't allow for easy refactoring.
2. add support for the rest of the submit arguments and the info attributes, and properly handle multi-cluster LSF
3. maybe move helper methods into a separate helper class (like the parsing methods)
4. possibly add some warnings 

The adapter as it stands allows submitting jobs through My Jobs and viewing jobs in Active Jobs for LSF 8.3.

PR #13, opened 2017-04-05
New job adapter API

### LINKED ISSUE #12: Split Adapter#info into separate methods?
We have one method that has 2 different ways of executing (one with id specified and one without) and two different return types:

1. if id is specified, implement algorithm optimized for getting the info of one job, and return a hash
2. if id is not specified, implement an algorithm optimized for getting the info of all the jobs, and return an array

It seems like we should split these into two separate methods. I think we overlooked this because the underlying implementation uses a single command (qstat, bjobs, squeue). 

Ideas: could be `info_all` and `info_find(id:)` or just `info_all` and `info(id:)`. That said, its kinda late in the game to make this change, it might be expensive.

nickjer:
Fixes #12

Also removes keyword args such as `script:` and `id:` as these are always required for the respective methods.

### DIFF
--- a/lib/ood_core/job/adapter.rb
+++ b/lib/ood_core/job/adapter.rb
@@ -8,11 +8,11 @@ module OodCore
       # @abstract Subclass is expected to implement {#submit}
       # @raise [NotImplementedError] if subclass did not define {#submit}
       # @example Submit job template to cluster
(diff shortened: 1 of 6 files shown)


nickjer:
Looks good.

Issue #27, opened 2017-05-04
LSF Adapter: add NodeInfo to Info using #procs for slots


ericfranz:
Notes from an email:

>Technically, LSF uses a construct called “job slots” which typically, as in our system, is configured to correspond to a core (although it needn’t necessarily). If a job runs on one job slot on one node, the execution hosts would be reported like “compute010”; two slots on a single node, it would go to “2*compute010” and on up to multiple nodes where a 48-slot job, for instance, running on all slots on three nodes would be reported as “16*compute010 16*compute011 16*compute012”. Mixed numbers like “5*compute010 11*compute011 16*compute012” are valid as well.

Issue #25, opened 2017-05-03
LSF Adapter: Add support all job submission options


ericfranz:
Most important:

- [x] 1. `Script#native`
- [x] 2. job_environment

Other Script parameters that look like they are supported:

- [x] start_time    `-b [[year:][month:]day:]hour:minute`
- [x] submit_as_hold `-H`
- [x] rerunnable `-r`
- [x] email `-u mail_user` 
    >   -u mail_user
    >               Sends mail to the specified email destination. To
    >               specify a Windows user account, include the domain
    >               name in uppercase letters and use a single
    >               backslash (DOMAIN_NAME\user_name) in
    >               a Windows command line or a double backslash
    >               (DOMAIN_NAME\\user_name) in a UNIX
    >               command line.

- [x] email_on_started `-B`
- [x] email_on_terminated `-N`
- [x] wall_time
    >-W [hour:]minute[/host_name |
    >/host_model]
    >            Sets the runtime limit of the batch job. If a UNIX
    >            job runs longer than the specified run limit, the
    >            job is sent a SIGUSR2 signal, and is killed if it
    >            does not terminate within ten minutes. If a
    >            Windows job runs longer than the specified run
    >            limit, it is killed immediately. (For a detailed
    >            description of how these jobs are killed, see
    >            bkill.)
    >

- [x] error_path `-e`
- [x] output_path `-o`
- [x] input_path `-i input_file` (supports %I and %J in input file name) or `-is`
- [x] priority `-sp priority` (integer 1 - MAX_USER_PRIORITY)
- [x] queue_name
- [x] reservation_id: `-U reservation_ID`

These don't look like they are supported

- args
- join_files
- min_phys_memory `-M [MB]` claims rosetta; bjobs man page for 8.3 says:
    >-M mem_limit
    > Sets a per-process (soft) memory limit for all the
    > processes that belong to this batch job (see
    > getrlimit(2)).
    > 
    > By default, the limit is specified in KB. Use
    > LSF_UNIT_FOR_LIMITS in lsf.conf to specify a
    > larger unit for the limit (MB, GB, TB, PB, or EB).

ericfranz:
https://slurm.schedmd.com/rosetta.pdf is useful

ericfranz:
The `Script#min_phys_memory` documentation states: 

>The minimum amount of physical memory in kilobyte

[...]

PR #28, opened 2017-05-08
Lsf Adapter Updates

### LINKED ISSUE #22: LSF Adapter: add "estimated runtime" by subtracting current time from start time
It is an estimate. If the job is never suspended after starting, this will be accurate.

LSF 9+ offers the ability to modify the output of the bjobs and specify runtime in the output. So we will be able to provide a more accurate runtime for later versions at that time.

This will allow an empty column in Active Jobs to be set with a value.

ericfranz:
Fixes #22 and #25 and #27

### DIFF
--- a/lib/ood_core/job/adapters/lsf.rb
+++ b/lib/ood_core/job/adapters/lsf.rb
@@ -72,24 +72,9 @@ module OodCore
           afternotok = Array(afternotok).map(&:to_s)
           afterany   = Array(afterany).map(&:to_s)
 
-          args = []
-          args += ["-P", script.accounting_id] unless script.accounting_id.nil?
-          args += ["-cwd", script.workdir.to_s] unless script.workdir.nil?
-          args += ["-J", script.job_name] unless script.job_name.nil?
+          kwargs = helper.batch_submit_args(script, after: after, afterok: afterok, afternotok: afternotok, afterany: afterany)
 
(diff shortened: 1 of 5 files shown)


nickjer:
Seems you are missing from the submission arguments `Script#email`

Quick question, does it really submit to the command line `-B true`, `-B false`, `-N true`, and `-N false` for the email on started and terminated?

nickjer:
Also I noticed this:

>-o output_file
>Specify a file path. Appends the standard output of the job to the specified file. Sends the output by mail if the file does not exist, or the system has trouble writing to it.
>
>-oo output_file
>Specify a file path. Overwrites the standard output of the job to the specified file if it exists, or sends the output to a new file if it does not exist. Sends the output by mail if the system has trouble writing to the file.

Do we want to append to the output file or write over it if it exists? I am assuming Torque and Slurm write over the output file it exists, but now that I think about it I never tested it.

nickjer:
Nevermind, it seems Torque actually appends output files as well. Hah, I never knew that.

ericfranz:
Good find on the email and email flags. I'll fix that.

ericfranz:
@nickjer final chance to review this before I merge it!

nickjer:
Sure, go ahead. If anymore bugs show up we can just patch them.

PR #37, opened 2017-05-25
[WIP] Add batch connect support to ood_core

nickjer:


nickjer:
@ericfranz: We have a small problem. The batch connect scripts aren't cluster specific. They are node specific. Meaning they depend on the software that is installed on any given node.

Typically the clusters all have the same software, but the Quick cluster can have different versions of software depending on the node you choose. So a cluster config such as:

```yaml
# /opt/ood/config/clusters.d/quick.yml
---

...
  batch_connect:
    vnc:
      vnc_env: |
        module load turbovnc/2.0.91
        export PATH=/usr/local/novnc/utils:${PATH}
```

May only work if the node has the given module specified.

nickjer:
I guess we could get fancy for the Quick cluster and have a script parse what the available turbovnc modules are and choose from a recommended ordered-list.

nickjer:
Other than feature detection, about option is asking systems to install a `ondemand/vnc` module on all nodes that sets up environment for approved turbovnc and websockify.

PR #43, opened 2017-06-27
Add helper method to filter jobs by owner

### LINKED ISSUE #26: Add an Job::Adapter#info_where as alternative to info_all
An example: `Adapter#info_where(user: "efranz")` or maybe better is `Adapter#info_where_user("efranz")`. We could also add `Adapter#info_where_queue("debug")` etc.

The `Adapter` superclass can offer a default implementation, which does a select on the results from `info_all`.

Individual adapters can optionally override with an optimized implementation. For example, bjobs by default shows only the user's jobs.

It could also be a `Adapter#info_where` filter and accept a hash where keys are methods on `Job::Info`.

nickjer:
Fixes #26

Changed it from `#info_where_user` to `#info_where_owner` because we defined it in the `Info` object as `#job_owner`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Features:
   - added a new Batch Connect template feature that builds batch scripts to
     launch web servers
   - removed deprecated tests for the Slurm adapter
+  - added method to filter list of batch jobs for a given owner
 
 Bugfixes:
 
--- a/lib/ood_core/job/adapter.rb
+++ b/lib/ood_core/job/adapter.rb
@@ -39,6 +39,13 @@ module OodCore
         raise NotImplementedError, "subclass did not define #info_all"
       end
 
+      # Retrieve info for all jobs for a given owner from the resource manager
+      # @param owner [#to_s] the owner of the jobs
+      # @return [Array<Info>] information describing submitted jobs
+      def info_where_owner(owner)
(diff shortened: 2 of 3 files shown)


brianmcmichael:
Is the goal here only to provide a convenience method for a user to request jobs?

I notice that this is still just requesting all jobs and then filtering down by the user, so it's not gaining much in the way of effienciency on systems where this call to all_jobs is slower. I guess each adapter would need to have it's own mechanism for requesting jobs by a specific user.

nickjer:
> Is the goal here only to provide a convenience method for a user to request jobs?

Not really, the goal is to provide a more optimized method to retrieve all jobs for a given owner. This is just the generic interface that will work for all adapters. The adapters themselves can then choose to override this with a more optimized routine.

brianmcmichael:
> The adapters themselves can then choose to override this with a more optimized routine.

Gotcha

nickjer:
An example:

```sh
┌─[jnicklas@puddle][~/Development/ood_core][±][pbspro ?:1 ✗]
└─▪ time qstat -f -t > /dev/null

real	0m33.129s
user	0m32.968s
sys	0m0.124s
┌─[jnicklas@puddle][~/Development/ood_core][±][pbspro ?:1 ✗]
└─▪ time qstat -u jnicklas > /dev/null

real	0m0.022s
user	0m0.007s
sys	0m0.008s
```

PR #46, opened 2017-06-29
Provide node info if available for non-running jobs in Slurm

nickjer:


ericfranz:
I'm uncomfortable with this change now. Can you imagine a situation where this would come back and bite us?

ericfranz:
For example, it seems like the more appropriate solution would be to add to the NodeInfo object scheduled_nodes alongside allocated_nodes and if an adapter can provide both it does.

nickjer:
It will only use `scheduled_nodes` if it exists and if no `allocated_nodes` are available (i.e., the job isn't running yet, most likely it is queued). Slurm doesn't provide both for a single job, if one exists the other doesn't.

I am not sure of the probability that those `scheduled_nodes` will be equivalent to the `allocated_nodes` when the job runs though. But I feel it can't hurt to provide more info for queued jobs, and let the app decide if it wants to use that info for a queued job.

nickjer:
The alternative is I just pad `allocated_nodes` as I am doing with PBS/Torque and PBS Pro with empty names for the nodes. I figured, since I had the names of the nodes in Slurm available to me, might as well just throw them in there and let apps decide if they want to use those names.

If you feel they should be empty strings instead, won't be that hard to remove the `scheduled_nodes` from this commit.

PR #47, opened 2017-06-30
Add optimized Adapter#info_where_owner code to PBS Pro adapter

nickjer:


nickjer:
@ericfranz: I made a class variable `PBSPro.factor` that decides the percentage at which it would be better to query the whole cluster and filter out the user's jobs.

Not sure if you would want this as an option on `PBSPro#info_where_owners(owners, factor: 0.10)`. I prefer this, but wasn't sure how you would feel about adapter-specific options on the interface methods.

PR #50, opened 2017-07-07
[WIP] LSF Adpater: info_where_owner optimization

ericfranz:
Implement support for optimized info_where_owner in LSF adapter.

Unfortunately, it doesn't seem that, in at least LSF 8.3, bjobs can accept a list of users as arguments. Either one user or one group or all users.

### DIFF
--- a/lib/ood_core/job/adapters/lsf.rb
+++ b/lib/ood_core/job/adapters/lsf.rb
@@ -19,6 +19,8 @@ module OodCore
 
     module Adapters
       class Lsf < Adapter
+        using Refinements::ArrayExtensions
+
         # @api private
         attr_reader :batch, :helper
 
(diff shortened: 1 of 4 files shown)


nickjer:
Maybe just fallback to `super` if `owner` is an array.

brianmcmichael:
The tests check out and I don't see any glaring errors here.

brianmcmichael:
Although, [this](https://github.com/OSC/ood_core/pull/50/files#diff-da5f385e23c574e13a73d454be9c48aaR116) seems redundant if `#wrap` returns an [empty array](https://github.com/OSC/ood_core/blob/master/lib/ood_core/refinements/array_extensions.rb#L12). I'm not familiar enough with the LSF output to know if you'd ever be in an alternative situation, however.

ericfranz:
Ah. Wrap returns an empty array. But I'm wrapping the array of owners. Returning `[]` if owners == `[]` makes sense in that the value being returned is not owners but is an array of JobInfo objects. No jobs for no owners :-).

PR #51, opened 2017-07-12
Remove `v1` cluster config support

### LINKED ISSUE #38: Deprecate `v1` backwards compatibility?
I feel we can safely deprecate the following code:

https://github.com/OSC/ood_core/blob/66524f83b9b257ac186bea416ea43f9e0a2ec8a4/lib/ood_core/clusters.rb#L47-L94

as all HPC centers that I worked with installing OOD uses the new `v2` cluster config.

nickjer:
Fixes #38 

The new format `v2` is now used at every OOD site installation. And all documentation has been updated to show `v2` formatting. Also the `v1` support previously was fairly neutered, and lacking many of the features (e.g., job adapter support) that are now required to run OOD apps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   manipulated through the `set_host` initialization parameter.
   [#42](https://github.com/OSC/ood_core/issues/42)
 
+### Removed
+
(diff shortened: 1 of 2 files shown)


ericfranz:
Ha this isn't deprecating, this is removing...

nickjer:
You are right. I took the leap from deprecation to removal in this PR.

Since all OOD installations now include `v2` formatting, I decided to take the leap.

PR #62, opened 2017-10-10
Fix bug when parsing nodes in Slurm

### LINKED ISSUE #54: Slurm node list
After using TACC I noticed a new format that the node list can come in:

```
c427-032,c429-002
```

I do not believe this is covered by the current Slurm adapter. In fact, I need to test the following formats:

```
c457-[011-012]
c439-021,c450-033
c439-[121-122]
c438-[062,104]
c433-[011,013]
c438-[052-053]
c431-[012,072]
c427-032,c429-002
c410-102,c414-004
c457-[001-002]
c474-[004,022]
c452-[054,121]
c453-[101,112]
c454-[021,064]
```

nickjer:
Fixes #54


ericfranz:
Why is this important? Is it because of the Ganglia feature where we show graphs for individual nodes?

nickjer:
> Why is this important?

It is a bug that causes `OodCore::Info#allocated_nodes` to give incorrect results for Slurm. So any apps that rely on that method may show incorrect results or crash depending on how they use that data.

This method appears both in ActiveJobs and the Dashboard.

PR #76, opened 2018-01-26
Fix for LSF adapter sometimes returning nil for job info

### LINKED ISSUE #75: LSF adapter returns `nil` instead of `Info` object
The `Adapter#info` method should only return an `Info` object and not `nil`. The LSF adapter returns `nil` here:

https://github.com/OSC/ood_core/blob/b80a6ee32a0cab6e306e8fefe8e85f6bf775098a/lib/ood_core/job/adapters/lsf.rb#L92

nickjer:
Fixes #75

### DIFF
--- a/lib/ood_core/job/adapters/lsf.rb
+++ b/lib/ood_core/job/adapters/lsf.rb
@@ -89,7 +89,14 @@ module OodCore
         def info(id)
           # TODO: handle job arrays
           job = batch.get_job(id: id)
-          job ? info_for_batch_hash(job) : nil
+          if job
+            info_for_batch_hash(job)
+          else
+            Info.new(
+              id: id,
+              status: :completed
+            )
+          end
         rescue Batch::Error => e
           raise JobAdapterError, e.message
         end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Fixed bug where LSF adapter would sometimes return `nil` when getting job
+  info. [#75](https://github.com/OSC/ood_core/issues/75)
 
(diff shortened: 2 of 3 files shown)


lib/ood_core/job/adapters/lsf.rb
```diff
@@ -89,7 +89,14 @@ def submit(script, after: [], afterok: [], afternotok: [], afterany: [])
...
           # TODO: handle job arrays
           job = batch.get_job(id: id)
-          job ? info_for_batch_hash(job) : nil
+          if job
+            info_for_batch_hash(job)
+          else
+            Info.new(
+              id: id,
+              status: :completed
+            )
```
ericfranz: Is it guarenteed that if job is nil the job is always completed and that other error cases we properly catch and raise? I'm assuming this is similar to how Torque works and is catching the case where the job is no longer in the queue.

Note it appears the other implementations of `#info` could possibly return nil as they all use detect which can return nil if not found. Though I didn't look further into those implementations to see.

nickjer: The job will only be set `nil` if `bjobs` returns an empty string and a successful exit code. Unfortunately, I am not sure how to exhaustively test this.

Some example cases:

1. Job that completed:

   ```console
   $ bjobs -a -w -W 1 2>/dev/null
   $ echo $?
   0
   ```

2. Job that was never created:

   ```console
   $ bjobs -a -w -W 1000000 2>/dev/null
   $ echo $?
   0
   ```

3. A job id that is impossible to exist:

   ```console
   $ bjobs -a -w -W 1000000000000000 2>/dev/null
   $ echo $?
   255
   ```

   ```console
   $ bjobs -a -w -W -1 2>/dev/null
   $ echo $?
   255
   ```

   ```console
   $ bjobs -a -w -W 10sdoijhf9432 2>/dev/null
   $ echo $?
   255
   ```

The first two test cases would return a completed job, whereas the "impossible" job id test cases would raise an error. I think that is reasonable.

I need to look at the other `#info` methods.

nickjer: All the other `#info` methods look good.

PR #83, opened 2018-03-30
Add shell path as script option

### LINKED ISSUE #82: Support shell path in Job::Script
It seems we have a good enough grasp on this option for the various adapters:

**Torque**:

```ruby
Job::Script.new(native: { headers: { Shell_Path_List: "/bin/bash" } })
```

**PBS Pro**

```ruby
Job::Script.new(native: ["-S", "/bin/bash"])
```

**Slurm**

Prepend `#!/bin/bash` to `Job::Script#content`

**LSF**

```ruby
Job::Script.new(native: ["-L", "/bin/bash"])
```

That it warrants adding this as an option.

nickjer:
Fixes #82

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Basic multi-cluster support for LSF by specifying name of cluster for -m
   argument. [#24](https://github.com/OSC/ood_core/issues/24)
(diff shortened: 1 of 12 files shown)


nickjer:
I have verified it works in the Dashboard using OSC's Torque. It should work everywhere else as I use the same arguments in the Dashboard. But I want to check a Slurm site just to be sure since it prepends to the script.

nickjer:
I have confirmed it works with PSC's Slurm scheduler.

Issue #93, opened 2018-08-09
Add ability to change VNC password length


Robbe7730:
Right now, the password length for `change_passwd` in `lib/ood_core/batch_connect/templates/vnc.rb` is hardcoded to be 8 characters. It would be nice to be able to change the length, possibly using the default value of `create_passwd`.

ericfranz:
@MorganRodgers you may have seen this already but https://github.com/OSC/ood_core/blob/d21a1d62a41b1a3e6d05a28cfe1100bb931597f9/lib/ood_core/batch_connect/templates/vnc.rb#L74-L75 is the offending code and it seems like if it would fetch the value from the context and use 8 as default otherwise then it would allow batch connect apps to change the value in the submit.yml (in the batch_connect hash).

ericfranz:
Though there definitely may be another better/preferred solution.

Robbe7730:
#94 fixes it, closing this issue. Thanks for the help!

ericfranz:
@Robbe7730 you are welcome! Thank you for opening the issue.

Issue #91, opened 2018-08-02
Method of generating password causes problems if compute node locale is not UTF8


ericfranz:
See https://listsprd.osu.edu/pipermail/ood-users/2018-August/000206.html

Perhaps we should use a different approach that would reliably work, even when the locale on the compute nodes is en_US instead of en_US.UTF-8

ericfranz:
The code:

https://github.com/OSC/ood_core/blob/d21a1d62a41b1a3e6d05a28cfe1100bb931597f9/lib/ood_core/batch_connect/template.rb#L145-L148 

Alternatives? Perhaps instead of `tr -cd '[:alnum:]'` we should be more explicit and do `tr -cd 'A-Z-a-z-0-9'`.

ericfranz:
An easy way to test would be to find offending characters, put it in a file and then do:

```
tr -cd '[:alnum:]' < /path/to/big/file/with/offending/characters 2> /dev/null | head -c${1:-#{passwd_size}} 
```

Then we can experiment/test/verify the difference. Or the create_passwd function accepts an optional argument and the file used in the function is either /dev/urandom or the file specified by the argument.

ke3vin:
You can also experiment  just by setting LANG (or LC_CTYPE) prior to running your test.
It appears that with LANG set to en_US, the `[:alnum:]` argument to `tr` generates a bunch of characters with their high bit set, I'm not sure why.  If LANG is set to en_US.UTF_8, it works correctly.
Using `tr -cd 'A-Za-z0-9'` works correctly in both cases.

ericfranz:
#94 uses `tr -cd 'A-Za-z0-9'` to fix the problem. Thank you @ke3vin for the help!

PR #94, opened 2018-08-20
Password context

MorganRodgers:
This sets a default password length of 8 to maintain backwards compatibilty, while providing the ability to override the default in `submit.yml.erb` by setting the key `batch_connect.passwd_size` to a positive non-zero integer value.

```yaml
---
batch_connect:
  template: vnc
  passwd_size: 16  # <-- optional new stuff
script:
  native:
    resources:
      nodes: "<%= bc_num_slots %><%= node_type %>"
```

Also fixes a bug where some values for the `LANG` environment variable might result in invalid passwords by making the character ranges available to `tr` more explicit.

Addresses #93 and #91.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Added
 - Added ability to override the default password length
 
-### Changed
-- Changed the default password length for VNC sessions to be 8
-
 ### Fixed
-- Fixed a bug in password creation where certain locales resulted in invalid passwords
+- Fixed a bug in password creation where certain locales resulted in invalid passwords [#91](https://github.com/OSC/ood_core/issues/91)
 
 ## [0.5.1] - 2018-05-14
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Added ability to override the default password length
+
+### Changed
+- Changed the default password length for VNC sessions to be 8
+
+### Fixed
+- Fixed a bug in password creation where certain locales resulted in invalid passwords
 
 ## [0.5.1] - 2018-05-14
 ### Fixed
--- a/lib/ood_core/batch_connect/template.rb
+++ b/lib/ood_core/batch_connect/template.rb
@@ -8,6 +8,8 @@ module OodCore
       using Refinements::HashExtensions
       using Refinements::ArrayExtensions
 
+      DEFAULT_PASSWD_SIZE = 32
+
       # The context used to render this template
       # @return [Hash] context hash
       attr_reader :context
@@ -68,6 +70,10 @@ module OodCore
         EOT
       end
 
+      def passwd_size
+        context.fetch(:passwd_size, DEFAULT_PASSWD_SIZE).to_i
+      end
+
       private
         # Working directory that batch script runs in
         def work_dir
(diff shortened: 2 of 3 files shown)


lib/ood_core/batch_connect/template.rb
```diff
@@ -68,6 +70,10 @@ def to_s
         EOT
       end
 
+      def passwd_size
```
ericfranz: It is unfortunate that the original function was titled `create_passwd` instead of `create_password` but IMHO we shouldn't propagate the abbreviation everywhere. Its preferable to use `password_size` i.e. `def password_size` and `DEFAULT_PASSWORD_SIZE` and `batch_connect.password_size`.

lib/ood_core/batch_connect/template.rb
```diff
@@ -68,6 +70,10 @@ def to_s
...
+        context.fetch(:passwd_size, DEFAULT_PASSWD_SIZE).to_i
```
ericfranz: Typically it is best practice to do as you have done: replace magic literals with a constant that describes what it is. However, in the case of Ruby code where we have attr accessors or methods that return a custom value OR a default value, it is clear what the magic literal is and so I'm not sure that taking the extra step to add the does not add more clarity.

These are examples of code where replacing with a constant doesn't add anything IMHO:

```ruby
# 8 is obviously the default password size
context.fetch(:passwd_size, 8).to_i

# 2000 is obviously default_min_port
min_port    = context.fetch(:min_port, 2000).to_i

# "/etc/ood/config/apps/dashboard" is obviously the default config_root
def config_root
  Pathname.new(ENV["OOD_APP_CONFIG_ROOT"] || "/etc/ood/config/apps/dashboard")
end

# "development" is obviously the default rails_env
def rails_env
  ENV['RAILS_ENV'] || ENV['RACK_ENV'] || "development"
end
```

But obviously in cases where it isn't clear, like when the magic literal is an operand in an expression, it definitely calls for it. Also, perhaps when the same literal is used in multiple places.

lib/ood_core/batch_connect/template.rb
```diff
@@ -95,7 +101,7 @@ def bash_helpers
           context.fetch(:bash_helpers) do
             min_port    = context.fetch(:min_port, 2000).to_i
             max_port    = context.fetch(:max_port, 65535).to_i
-            passwd_size = context.fetch(:passwd_size, 32).to_i
+            # passwd_size = context.fetch(:passwd_size, 32).to_i
```
ericfranz: Remove the comment since its no longer used.

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Added ability to override the default password length
+
+### Changed
+- Changed the default password length for VNC sessions to be 8
```
ericfranz: The default password length used to be 8, so I'd just remove the changed section here.

CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
...
+
+### Fixed
+- Fixed a bug in password creation where certain locales resulted in invalid passwords
```
ericfranz: Add a link to the github issue that is fixed i.e. `[#91](https://github.com/OSC/ood_core/issues/91)`

lib/ood_core/batch_connect/template.rb
```diff
@@ -95,7 +101,7 @@ def bash_helpers
           context.fetch(:bash_helpers) do
             min_port    = context.fetch(:min_port, 2000).to_i
             max_port    = context.fetch(:max_port, 65535).to_i
```
ericfranz: You pulled out `passwd_size` into its own method, which is nice. Would it improve code readability to do the same with min_port and max_port?

PR #100, opened 2018-10-26
Sge

MorganRodgers:
The current state of the Open Ondemand (Sun|Son of|Univa) Grid Engine adapter.


MorganRodgers:
During a load test of the adapter the gridengine-master process crashed. DRMAA seems like it might be required.

lib/ood_core/job/adapters/sge/batch.rb
```diff
@@ -0,0 +1,194 @@
...
+      DRMAA::STATE_DONE                  => :completed,
+      DRMAA::STATE_FAILED                => :completed
+    }
+
+    # The one and only connection with DRMAA
+    # Attempting to instantiate a DRMAA::Session more than once causes it to crash
+    class SessionSingleton < DRMAA::Session
+      include Singleton
+    end
+  rescue LoadError; end
```
ericfranz: Should we log a warning here? Or is it expected functionality that this load attempt happens multiple times in the normal execution?

If we want to log a warning we could open an issue to add a ood_core wide logger. There might be an issue already open. Something where an app can set the wide logger to its logger i.e. `OodCore.logger = Rails.logger`.

lib/ood_core/job/adapters/sge/batch.rb
```diff
@@ -0,0 +1,194 @@
...
+
+  # @param opts [#to_h] the options defining this adapter
+  # @option opts [Batch] :batch The Sge batch object
+  #
+  # @api private
+  # @see Factory.build_sge
+  def initialize(config)
+    @cluster          = config.fetch(:cluster, nil)
+    @conf             = Pathname.new(config.fetch(:conf, nil))
+    @begin            = Pathname.new(config.fetch(:bin, nil))
```
ericfranz: Change to bin

lib/ood_core/job/adapters/sge/batch.rb
```diff
@@ -0,0 +1,194 @@
...
+  # @return job_id [String]
+  def submit(content, args)
+      cmd = ['qsub'] + args
+      @helper.parse_job_id_from_qsub(call(*cmd, :stdin => content))
+  end
+
+  # Call a forked SGE command for a given batch server
+  def call(cmd, *args, env: {}, stdin: "", chdir: nil)
+    cmd = cmd.to_s
+    cmd = @bin.join(cmd).to_s if @bin
```
ericfranz: This will always run because bin can never be nil

ericfranz: - `@helper` `@bin` replace with getters `attr_reader`

lib/ood_core/job/adapters/sge/batch.rb
ericfranz: Since you are considering fixing the tests to mock Open3 maybe verify or update tests to catch this error before fixing it.

lib/ood_core/job/adapters/sge/batch.rb
```diff
@@ -0,0 +1,194 @@
...
+      listener = QstatXmlJRListener.new
+      argv = ['qstat', '-r', '-xml', '-j', job_id.to_s]
+      REXML::Parsers::StreamParser.new(call(*argv), listener).parse
+
+      job_hash = listener.parsed_job
+      job_hash[:status] = get_status_from_drmma(job_id) if can_use_drmaa?
+        
+      job_info = OodCore::Job::Info.new(**job_hash)
+    rescue Error, REXML::ParseException  # Job not found
+    end
```
ericfranz: Why rescue Error here? In this case `call(*argv)` can result in an error, if for example, the scheduler is unavailable, or OnDemand has been installed on a new node and that node hasn't been added to the list of hosts SGE will accept qstat calls from, etc.

In the error case I would expect the error to bubble up. Instead it will be caught, ignored, and the job will be reported as completed.

Also, would a REXML::ParseException occur if the output of the `QstatXmlJRListener` doesn't handle the output properly say in the case that the XML is missing a tag or the format is different? Or is the REXML::ParseException only to ever occur if the xml is well formatted but the job id is not returned by the call?

Oh perhaps when the job id doesn't exist the qstat call returns an exit status not 0?

lib/ood_core/job/adapters/sge/batch.rb
```diff
@@ -0,0 +1,194 @@
...
+
+    job_info
+  end
+
+  def can_use_drmaa?
+    @sge_root && Object.const_defined?('DRMAA')
```
ericfranz: Use getter `sge_root` instead of instance var directly

lib/ood_core/job/adapters/sge/qstat_xml_j_r_listener.rb
```diff
@@ -0,0 +1,115 @@
+require 'rexml/document'
+require 'rexml/streamlistener'
+require 'date'
+
+# An XML stream listener to build an array of OodCore::Job::Info from qstat output
+#
+# Handles parsing `qstat -xml -r` which provides:
```
ericfranz: Fix comment as this handles parsing of `qstat -xml -j -r`. :-) copy paste woes

spec/job/adapters/sge/batch_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+  let(:job_from_qacct) {
+    OodCore::Job::Info.new(
+      :accounting_id => nil,
+      :allocated_nodes => [],
+      :cpu_time => nil,
+      :dispatch_time => Time.parse('2018-10-25 13:16:29 +0000'),
+      :id => "1072",
+      :job_name => "job_7",
+      :job_owner => "vagrant",
+      :native => nil,
```
ericfranz: In the other adapters we have returned the object representation of the original set of attributes returned from the adapter in this native hash. 

Since you aren't parsing the output of qstat using a parse-able column output, but rather XML, I can see why it might be preferable only to capture specific values. But it might be worth at least adding to native the sqe state.

The full native hash might look like this, given this XML:

```xml
    <job_list state="pending">
      <JB_job_number>1045</JB_job_number>
      <JAT_prio>0.25000</JAT_prio>
      <JB_name>job_RQ</JB_name>
      <JB_owner>vagrant</JB_owner>
      <JB_project>project_b</JB_project>
      <state>qw</state>
      <JB_submission_time>2018-10-09T18:47:05</JB_submission_time>
      <slots>1</slots>
      <full_job_name>job_RQ</full_job_name>
      <hard_request name="h_rt" resource_contribution="0.000000">360</hard_request>
      <hard_req_queue>general.q</hard_req_queue>
    </job_list> 
```

to

```ruby
{
  JB_job_number: 1045,
  JAT_prio: 0.25000,
  JB_name: "job_RQ",
  JB_owner: "vagrant",
  JB_project: "project_b",
  state: "qw",
  JB_submission_time: "2018-10-09T18:47:05",
  slots: 1,
  full_job_name: "job_RQ",
  hard_request: 360,
  hard_request_name: "h_rt",
  hard_request_resource_contribution: "0.000000",
  hard_req_queue: "general.q"
}
```

And we would add an `extended_data_sge` method to Active Jobs that contained this SGE specific information. Here is the Slurm example: https://github.com/OSC/ood-activejobs/blob/b77a6889b61fe03af13902bcb0c0cccada2530da/app/models/jobstatusdata.rb#L101-L128/jobstatusdata.rb#L101-L128

And that is what appears in the table when you expand the details of a job in the Active Jobs app.

Note: I think it is worth stating that sending all this extra information in the native hash costs something, and the default information for Job::Info might be good enough for many sites, or only a subset of the native hash might be good enough, and thus for these reasons it might be preferable to have a mode where adapters do not set the native hash on the Job::Info object. That is one consideration for the performance improvements after this OOD release.

ericfranz: Also, removing adapter specific knowledge from ActiveJobs, which I think can be done.

ericfranz: Two benefits of a native hash:

1. Users can see adapter specific information that may not have a corresponding attr in the Info object
2. Sometimes we parse a value into a generic format and that parsing is incorrect or strips valuable information. The original value can be accessed by the user if that value is in the native hash

ericfranz: While we don't do it in other adapters yet, we could omit from the native hash any values that are identical duplicates on job info attrs. That would save a little memory/time.

ericfranz: Or maybe for the sake of time, just return {} for the native hash, write up an issue to add extended data to SGE, we can then add this in 1.5 release after we do performance improvements.

lib/ood_core/job/adapters/sge/batch.rb
ericfranz: Change to load in a `self.load` that is executed conditionally if sge_root is defined AND raise an exception if there is a LoadError or let LoadError bubble up.

lib/ood_core/job/adapters/sge/helper.rb
```diff
@@ -0,0 +1,65 @@
...
+    raise OodCore::Job::Adapters::Sge::Error.new('SGE does not support job dependencies on after not ok') if afternotok && ! afternotok.empty?
+    raise OodCore::Job::Adapters::Sge::Error.new('SGE does not support job dependencies on after any') if afterany && ! afterany.empty?
+  end
+
+  # Extract the job id from qsub's output
+  # e.g. Your job 1043 ("job_16") has been submitted
+  # @param qsub_output [#to_s]
+  # @return job_id [String]
+  def parse_job_id_from_qsub(qsub_output)
+    /Your job (?<job_id>[0-9]+)/.match(qsub_output)[:job_id]
```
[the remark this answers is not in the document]
ericfranz: Maybe fix it from the job array branch to your merge conflict chagrin (or pull from the job array PR)

MorganRodgers:
@ericfranz I had forgotten that I had merged the working state of job_arrays into my development SGE branch. I want to remove the job_array stuff while still fixing the bug with SGE job ID parsing.

spec/job/adapters/sge/batch_spec.rb
MorganRodgers: Opened #114.

MorganRodgers: Updated `#info` and `#info_all` to always return an empty hash for the time being.

lib/ood_core/job/adapters/sge/batch.rb
MorganRodgers: I have removed the catch on `Error`. As for `ParseException` when `qstat -xml -j` fails to find a job it responds with invalid XML...because design? I confirmed this confusing behavior on both my development VM and UCLA's Hoffman cluster so it appears reliable.

ericfranz: I guess one final risk here is the assumption that results will contain a newline :-)

```
results.split("\n")[1]
```

What about just a regex:

```ruby
raise e unless results =~ /unknown_jobs/
```

This actually also works if results happened to be nil for whatever reason.

ericfranz: Is it worth customizing the message, something like:

```ruby
raise e, "REXML::ParseException error and command '#{argv.join(' ')' produced results that didn't contain string 'unknown_jobs'. ParseException: #{e.message}"
```

lib/ood_core/job/adapters/sge/helper.rb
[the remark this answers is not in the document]
MorganRodgers: I copied it into here so we don't propagate a known bug. I'll deal with conflicts when I get back to job_arrays.

lib/ood_core/job/adapters/sge/batch.rb
MorganRodgers: I think that is probably the safest thing to do.

PR #104, opened 2018-11-09
WIP: Improved support for job arrays: TaskStatus and Info updates

MorganRodgers:
The current TODO list for Job Array support:

* [x] Add `OodCore::Job::TaskStatus` - (or ChildTaskStatus?)
* [x] Add `TaskStatus` to `OodCore::Job::Info`
* [x] Add status aggregation to `Info` when child task statuses are present
* [ ] Ensure that calls to `OodCore::Job::AdapterSubclass#submit` returns a valid input to `AdapterSubclass#info`
    * [ ] LSF
    * [ ] PBSPro
    * [ ] Slurm
    * [ ] Torque
    * [ ] SGE (coming soon?)
* [ ] Ensure that calls to `OodCore::Job::AdapterSubclass#info` pass the correct native flags if the job id is a job array parent, and that child job statuses are set on the resulting `Info`
    * [ ] LSF
    * [ ] PBSPro
    * [ ] Slurm
    * [ ] Torque
    * [ ] SGE (coming soon?)
* [ ] Ensure that calls to `OodCore::Job::AdapterSubclass#delete` pass the correct native flags if the job id is a job array parent
    * [ ] LSF
    * [ ] PBSPro
    * [ ] Slurm
    * [ ] Torque
    * [ ] SGE (coming soon?)


lib/ood_core/job/info.rb
```diff
@@ -120,7 +131,8 @@ def to_h
           cpu_time:        cpu_time,
           submission_time: submission_time,
           dispatch_time:   dispatch_time,
-          native:          native
+          native:          native,
+          child_task_statuses: child_task_statuses
```
ericfranz: Maybe we should rename this to be: `child_tasks` or just `tasks`. And right now a task only contains a status and an id, but it is a bucket for other possible relevant attributes in the future and then the method naming makes more sense to me: `OodCore::Job::Task#status`  - returns the status of an `OodCore::Job::Task` sounds better than `OodCore::Job::TaskStatus#status` returns the status of an `OodCore::Job::TaskStatus`

Though its true that is not what we previously discussed. Do you see any drawbacks to having a "Task" object over a "TaskStatus" object?

MorganRodgers: I'm good with the alternate naming.

lib/ood_core/job/adapters/torque.rb
```diff
@@ -201,7 +201,12 @@ def info_where_owner(owner)
         # @see Adapter#info
         def info(id)
           id = id.to_s
-          parse_job_info(*@pbs.get_job(id).flatten)
+
+          result = @pbs.get_job(id)
+
+          return parse_job_info(*result.flatten) if result.keys.length == 1
+
+          parse_job_array(id, result)
```
ericfranz: I thought that `@pbs.get_job(id)` would return either an array of hashes or a single Hash. And that it would return an Array of hashes for a job array. Thus the original implementation assumed a single job was being returned for `info(id)` call i.e. `info("10219837.oak-batch.osc.edu")` would return below:

```
irb(main):009:0> {"10219837.oak-batch.osc.edu" => { :Job_Name => "CFD_Solver" } }.flatten
=> ["10219837.oak-batch.osc.edu", {:Job_Name=>"CFD_Solver"}]
```

And then `*["10219837.oak-batch.osc.edu", {:Job_Name=>"CFD_Solver"}]` makes that two arguments, the id `"10219837.oak-batch.osc.edu"` and the hash `{:Job_Name=>"CFD_Solver"}`.

So I would expect this code to crash, as I would expect `get_job` to return an array if it were a job array, and arrays do not respond to `keys` method. Is this correct?

Also, is it a job array only if the array size is > 1? Or is there any situation where you have a job array where get_job returns only 1 job? If you reliably only get array size > 1 if there is a job array, then you can coerce result into an array using the refinement here https://github.com/OSC/ood_core/blob/d21a1d62a41b1a3e6d05a28cfe1100bb931597f9/lib/ood_core/refinements/array_extensions.rb

```ruby
result = Array.wrap(@pbs.get_job(id))
```

Then the info implementation is simplified to:

```ruby
def info(id)
  id = id.to_s
  result = Array.wrap(@pbs.get_job(id))
  parse_job_info(id, result.first.flatten.last, result.drop(1).map { |k,v| { id: k, status: STATE_MAP.fetch(value[:job_state], :undetermined) } })
```

Or I guess more readable:

```ruby
def child_task_statuses(child_hashes) # what is a good name for the arg child_hashes?
  child_hashes.map do |key, value|
    id: k, status: STATE_MAP.fetch(value[:job_state], :undetermined) } }
  end
end

def info(id)
  id = id.to_s
  result = Array.wrap(@pbs.get_job(id))
  parse_job_info(id, result.first.flatten.last, child_task_statuses(result.drop(1)))

File: lib/ood_core/job/adapters/slurm.rb [hunk 3]
```diff
@@ -334,13 +334,24 @@ def info(id)
...
-          # array id, so we need to find the job that corresponds to the
-          # given job id (if we can't find it, we assume it has completed)
-          info_ary.detect( -> { Info.new(id: id, status: :completed) } ) do |info|
-            # Match the job id or the formatted job & task id "1234_0"
-            info.id == id || info.native[:array_job_task_id] == id
+          # If no job was found we assume that it has completed
+          return Info.new(id: id, status: :completed) unless ! info_ary.empty?
+
+          # If only one job was returned we return it
+          return info_ary.first unless info_ary.length > 1
```
ericfranz: See discussion on Torque adapter below. IMHO multiple returns in a method body like this is a bad idea if it can be replaced with a functional equivalent and only one return statement.

lib/ood_core/job/adapters/torque.rb
ericfranz: So the assumptions in this code and `parse_job_array` as I understand it are:

1. `@pbs.get_job(id)` returns a hash of key value pairs, one for each job or task. So a regular job only will have one key value pair, and a job array will have multiple
2. If we are dealing with a job array, the parent task is the first hash, both returned by `result.keys.first` and `result.to_a.first`. Is this guaranteed to always be the case, or is there a better way to determine the parent task than placement in the hash (such as the id - is the id of the parent task more closely related to the id used when calling `get_job`?
3. I also noticed `parent_task.delete(:exec_host)`. It says "Master tasks don't actually run on a host" but child tasks do? The problem I see here is that for child tasks we are not preserving exec_host, only the task id and statuses. I thought the entire job array ran on the same exec_hosts or at least in the same set of nodes allocated for the job. So I imagine we would want to preserve this information when displaying job info for a single job. Could you explain what you were thinking here?

ericfranz: Oh I was mistaken, from reading the code it would seem that for a job array get_jobs returns one hash per task. So the "parent task is just getting the first task's hash, and the assumption is the rest of the values are shared.

So does that mean that parent_task has a status equal to the value of the status of the first task?

MorganRodgers: That's correct that the first hash returned is used to populate most values for the parent task. That may not be what we want. If the first job has begun executing then it may have CPU time which isn't really valid for a 'parent' job.

As for the status of a parent task: no. If there are `#tasks` then `job_array_aggregate_task_status` is called against the statuses of all child tasks.

lib/ood_core/job/task.rb
```diff
@@ -0,0 +1,24 @@
...
+
+      def initialize(id:, status:, **_)
+        @task_id = id
+        @status = OodCore::Job::Status.new(state: status)
+      end
+
+      def to_h
+        {
+          :id => @id,
+          :status => @status
```
ericfranz: best practice to use the getters here instead of the instance vars directly:

```ruby
def to_h
  {
    :id => id,
    :status => status
  }
end
```

lib/ood_core/job/adapters/torque.rb
ericfranz: So instead of `parent_task.delete(:exec_host)` I imagine we would be doing something like this:

```ruby
# exec host looks like "n0163/2,7,10-11+n0205/0-11+n0156/0-11"
# assuming exec host for a non running job is nil
# assuming exec host strings from multiple nodes can be merged with a +; but if each task is running on the same node the result may be silly: "n0163/0-27+n0163/0-27+n0163/0-27+n0163/0-27" or maybe slightly informative "n0163/0+n0163/1+n0163/2+n0163/3"
parent_task[:exec_host] = results.map { |k,v| v[:exec_host] }.compact.join("+")
```

The beneficial thing about this approach is that this code works to get the exec_host of a normal job.

For example in irb:

```
irb(main):001:0> job = {"4205673.owens-batch.ten.osc.edu" => { exec_host: "o0732/0" }}
=> {"4205673.owens-batch.ten.osc.edu"=>{:exec_host=>"o0732/0"}}
irb(main):002:0> jobary =  {"4205664[1].owens-batch.ten.osc.edu" => { exec_host: "o0732/0" }, "4205664[2].owens-batch.ten.osc.edu" => { exec_host: "o0732/1"}}
=> {"4205664[1].owens-batch.ten.osc.edu"=>{:exec_host=>"o0732/0"}, "4205664[2].owens-batch.ten.osc.edu"=>{:exec_host=>"o0732/1"}}
irb(main):003:0> def aggregate_exec_host(results)
irb(main):004:1>   results.map { |k,v| v[:exec_host] }.compact.join("+")
irb(main):005:1> end
=> nil
irb(main):007:0> aggregate_exec_host jobary
=> "o0732/0+o0732/1"
irb(main):006:0> aggregate_exec_host job
=> "o0732/0"
irb(main):009:0> job["4205673.owens-batch.ten.osc.edu"][:exec_host]
=> "o0732/0"
irb(main):011:0> job["4205673.owens-batch.ten.osc.edu"][:exec_host] == aggregate_exec_host(job)
=> true
```

ericfranz: If we take this approach, then two methods can be written, one to get the aggregate_exec_host and another to get the task_list and for a normal job aggregate_exec_host returns the exec_host and task_list returns [] and you won't need any special cases or control flow statements.

MorganRodgers: Added in 148f12f.

lib/ood_core/job/info.rb
```diff
@@ -143,6 +155,16 @@ def eql?(other)
...
+
+      private
+
+      # Generate an aggregate status from child tasks
+      # @return [OodCore::Job::Status]
+      def job_array_aggregate_status
+        @tasks
+         .map { |task_status| task_status.status}
+         .max
+      end
```
ericfranz: This is an example where since a single line is < 80 characters, I actually find the single line statement easier to read. Maybe its because I'm just used to seeing it that way?

```ruby
def job_array_aggregate_status
  @tasks.map { |task_status| task_status.status}.max
end
```

What style guide are you using?

spec/job/array_ids_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  context "when the spec is for multiple ranges" do
+    it "returns the correct IDs" do
+      expect( described_class.new('1-2,4-5,7-9').ids ).to eql([1, 2, 4, 5, 7, 8, 9])
+    end
+  end
+
+  context "when the spec contains a percent modifier" do
+    it "returns the correct IDs" do
+      expect( described_class.new('1-4%2').ids ).to eql([1, 2, 3, 4])
```
ericfranz: What is the percent modifier for? I notice `1-4%2` gets the same results as `1-4`.

spec/job/array_ids_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  context "when the spec is for a list of single ids" do
+    it "returns the correct IDs" do
+      expect( described_class.new('1,3,5,7,11').ids ).to eql([1, 3, 5, 7, 11])
+    end
+  end
+
+  context "when the spec is for multiple ranges" do
+    it "returns the correct IDs" do
+      expect( described_class.new('1-2,4-5,7-9').ids ).to eql([1, 2, 4, 5, 7, 8, 9])
```
ericfranz: I see multiple ranges, a list of single ids, and a compound range and single id, but what about a mixture

```
1,3-6:3,7,9-11,13,15-17,20-30:2
```

lib/ood_core/job/info.rb
MorganRodgers: I am not currently using a style guide. Most of my interest in one at the beginning turned out to be trying to understand all the things that can be done with Ruby.

spec/job/array_ids_spec.rb
MorganRodgers: The percent modifier is used by a Slurm and Torque as a limiter on the entire array job. In Slurm it controls the number of simultaneous tasks, in Torque it specifies a slot limit. In either case that information is not pertinent to what we are looking for, and so I discard it.

lib/ood_core/job/info.rb
ericfranz: I thought about it more and I can see the value in representing a chain vertically, especially when it gets longer. Also, the “complexity at a glance” might not be the best metric to grade code readability on - which is what I used in my initial comment.

lib/ood_core/job/adapters/sge/qstat_xml_j_r_listener.rb
```diff
@@ -112,5 +136,46 @@ def end_CE_stringval
...
+    @running_tasks << @current_text
+  end
+
+  def set_job_array_piece(key)
+    @job_array_spec[key] = @current_text if @processing_job_array_spec
+  end
+
+  def spec_string
+    '%{start}-%{stop}:%{step}' % @job_array_spec
+  end
```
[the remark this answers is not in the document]
ericfranz: Maybe default if malformed is `1-1:1`?

lib/ood_core/job/info.rb
```diff
@@ -79,12 +84,14 @@ class Info
       # @param cpu_time [#to_i, nil] cpu time
       # @param submission_time [#to_i, nil] submission time
       # @param dispatch_time [#to_i, nil] dispatch time
+      # @param tasks [Array<Hash<#to_s, #to_sym>>] tasks
```
ericfranz: Expand comment so its i.e. `{ id: "id", status: :status }` such as `{ id: '12345.owens-batch', status: :running }`.

lib/ood_core/job/status.rb
```diff
@@ -113,6 +115,25 @@ def hash
           self == state
         end
       end
+
+      # Determines whether this method corresponds to a status check for a valid
+      # state
+      # @param method_name the method name called
+      # @return [Boolean]
+      def respond_to_missing?(method_name, include_private = false)
```
ericfranz: remove this method

lib/ood_core/job/adapters/sge.rb
```diff
@@ -90,8 +90,8 @@ def submit(script, after: [], afterok: [], afternotok: [], afterany: [])
 
         # Retrieve info for all jobs from the resource manager
         # @return [Array<Info>] information describing submitted jobs
-        def info_all
-          @batch.get_all(owner: '*')
+        def info_all(attrs: nil)
```
ericfranz: Fix this

```
@batch.get_all(owner: '*')
```

MorganRodgers: Fixed.

Issue #89, opened 2018-07-20
Simplify Adapter loading code


ericfranz:
The adapter loading code is currently modeled off of how Rails loads database adapters. It is unnecessarily complex. It provides two advantages:

1. Each adapter can define the lowercase string representation of the class that can be used in the config file, and the factory knows what file to require and what method to execute.
2. Each adapter can have its own factory method to do extra setup before handling the configuration options in the initializer
3. Adapter class is only required when the config file demands it be loaded.

I don't see a reason why we can't just require all the classes. Since any class that uses native code does so via FFI, and since this code is executed only at runtime, there should be no problem requiring the TorqueAdapter when using an LSF system.

If our code requires all the adapters like normal, and each adapter accepts the config hash as an argument to the initializer, then we can just do this:

```ruby
def build(config)
  config = config.to_h.symbolize_keys
  adapter = config.fetch(:adapter) { raise AdapterNotSpecified, "job configuration does not specify adapter" }.to_s
  adapter_class = OodCore::Job::Adapters.constants.find { |cls| cls.to_s.underscore == adapter }
  raise AdapterNotFound, "job configuration specifies nonexistent #{adapter} adapter" if adapter_class.nil?
  adapter_class.new(config)
end
```

Less code in the factory method and each adapter.

ericfranz:
Actually this could be made much simpler. Just have a global hash whose key is the "adapter" and whose value is the "constant":

```ruby
module OodCore
  module Job
    class Factory
      class << self
        attr_accessor :adapters
        adapters = {
          "slurm" => OodCore::Job::Adapters::Slurm,
          "lsf" => OodCore::Job::Adapters::Lsf,
          "pbspro" => OodCore::Job::Adapters::PBSPro,
          "torque" => OodCore::Job::Adapters::Torque,
        }
      end

      def build(config)
        adapter = config.fetch(:adapter) { raise AdapterNotSpecified, "job configuration does not specify adapter" }.to_s
        adapter_class = self.class.adapters.fetch(adapter) { raise AdapterNotFound, "job configuration specifies nonexistent #{adapter} adapter" }
        adapter_class.new(config)

        # or to preserve existing functionality
        # adapter_class.build(config)
      end
    end
  end
end
```

ericfranz:
I guess the constants search would actually reduce one step when adding a new adapter... as you wouldn't have to modify this Hash. The Hash modification gives you more flexibility but I'm not sure we need that - convention over configuration.

[...]

PR #105, opened 2018-11-16
Fix issue with Slurm 18 dropping a format code

MorganRodgers:


ericfranz:
So according to https://slurm.schedmd.com/archive/slurm-17.11-latest/squeue.html `%b` or `gres` is "Generic resources (gres) required by the job or step. (Valid for jobs and job steps) ". I wonder why they removed this option (and could they remove other options in the future? what about adding new options?). Interestingly I don't think we actually use this value anywhere. I could be wrong though - can you confirm we don't actually report this value anywhere? For example, one option is we just remove this across the board - if it is an attribute only on the "native representation". Though I guess that would technically be an interface change requiring the major version change. So maybe your suggested solution is the best.

Assuming the the solution in this PR is the correct solution, I would do the following to avoid shelling out during initialization of the object and support optionally setting the version directly in the cluster config to avoid shelling out altogether.

```ruby
def initialize(cluster: nil, bin: nil, conf: nil, version: nil)
  @cluster = cluster && cluster.to_s
  @conf    = conf    && Pathname.new(conf.to_s)
  @bin     = Pathname.new(bin.to_s) 
  @version = version unless version.nil?
end

def version
  @version ||= detect_version
end
```

- if `@version` is not set, `detect_version` will be executed when `version` is first called and will set the value `@version`. Future calls to `version` will just return `@version`.

Also, assuming the order of the hash does not matter, you could do this as well:

```diff
def fields
  {
    account: "%a",
    job_id: "$A",
-   gres: "%b"
    ...
    work_dir: "%Z"
- }
+ }.tap { |f| f[:gres] = "%b" if version < 18 }
end
```

Or delete as before:

```diff
def fields
  {
    account: "%a",
    job_id: "$A",
    gres: "%b"
    ...
    work_dir: "%Z"
- }
+ }.tap { |f| f.delete(:gres) if version >= 18 }
end
```

- If you aren't familiar, `tap` is a useful method that executes the block with the object as the argument, and then returns the object (i.e. the hash in this case). So here we add the gres option for older versions of Slurm.

Finally in `detect_version` instead of using `@bin` use `bin`.

ericfranz:
So this changed would now be executing "detect_version" for all current Slurm sites. Maybe instead we just require version as a value and assume it is 17 if no version is set. Sites running Slurm 18 could be instructed to set the version in their config file. We could remove the code to auto-detect the version using `sbatch --version`.

MorganRodgers:
I don't think that the changes I added amount to too much of a performance hit (but have no data). On the chance that someone somewhere might want to use `gres` I think that removing it across the board isn't ideal.  I like the idea of adding a default parameter for version, and just requiring version 18 users to explicitly set the version in their configs. Likewise I think that I prefer requiring admins to set the version to attempting to auto-detect.

I did not know about `Object#tap`, but that looks useful.

ericfranz:
Instead of asking explicitly for "major version" I think that just "version" might be preferable. You can do this:

```ruby
major_version = c.fetch(:version, 17).to_i
major_version = 17 if major_version == 0 # or perhaps we should be raising an exception if an illegal value is put here instead of silently assuming v17
```

Based on behavior in interpreter:

```
➜  nginx_stage git:(refactor) ✗ irb
irb(main):001:0> "18.2.1".to_i
=> 18
irb(main):002:0> 18.to_i
=> 18
irb(main):003:0> "18".to_i
=> 18
irb(main):004:0> "v18".to_i
=> 0
irb(main):005:0> "version eighteen".to_i
=> 0
```

This keeps options open in the future if we have to add a fix for a minor or patch version, and other adapters can follow suit. More complex version comparisons can be accomplished using the gem library version string comparison methods. For example, I do think there is a significant difference between LSF 9.0 and 9.1 but of course the biggest difference is between 8.3 and 9.1.

I think the only reason I'd be suggesting this instead of the suggested solution is this is almost as easy to implement and keeps more options open. I'm not wanting to suggest we do extra work for a problem we don't have right now.

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -11,16 +11,26 @@ class Factory
...
         slurm = Adapters::Slurm::Batch.new(cluster: cluster, conf: conf, bin: bin, major_version: major_version)
         Adapters::Slurm.new(slurm: slurm)
       end
+
+      def parse_major_version(version)
+        # Backward compatible behavior for current installations
+        return 17 if version.nil?
+
+        version.to_i
+      end
```
ericfranz: Why do `major_version = parse_major_version(c.fetch(:version, nil))` when the same functionality can be done by just `major_version = c.fetch(:major_version, 17).to_i`?

ericfranz: Also note that this block of code is actually opening OodCore::Job::Factory and adding the method `build_slurm` and `parse_major_version`. At least `build_slurm` has somewhat of a "namespace" with the suffix, but the `parse_major_version` is probably too generic of a method to be added to OodCore::Job::Factory. Imagine if we took the same approach with another adapter wanted to implement a similar method - one implementation would replace the other introducing a defect because we would now have two implementations of `OodCore::Job::Factory#parse_major_version`.

ericfranz: See https://github.com/OSC/ood_core/issues/89 - I want to simplify the creation of these adapter instances. The result would be removing or moving the build methods to the Adapter classes themselves. 

If you are curious why the original complexity exists, we first copied what Rails does with loading database adapters since. But YAGNI.

ericfranz:
I can't help but wonder if we are doing this wrong. Every adapter is going to have this issue at some point. The output of all of these commands produces a top row that has the names of the columns. Perhaps we should be translating the column names to key values and then populating a hash, instead of using the hash to determine the expected columns (and thus crashing when one doesn't exist). Though perhaps a solution like that would be much more difficult, and asking admin's to specify a version # in a cluster config is not that difficult.

ericfranz:
>On the chance that someone somewhere might want to use gres I think that removing it across the board isn't ideal.

AFAIK this is in the native only and we don't map it to any current value in info, and this is not reported anywhere. The likelihood of someone making use of this is low. I think the pragmatic solution would be to remove gres. Then the adapter just works for 17 and 18. 

Technically a breaking change? I don't know. We don't actually define the interface if the native attrs anywhere.

treydock:
FWIW SLURM didn't drop `%b` from squeue output, it might have just changed formatting.  https://github.com/SchedMD/slurm/commit/bbcd61c4d0c942b57331a4c57af2579856e2142e#diff-cc338e582d660acb044a6e4c1c4e4cdd

It went from being GRES to being TRES, which is going from generic resources to trackable resource.  The man page for `squeue` does not mention `%b` but it's still in code and still can be used.

```
$ squeue --format '%b'
TRES_PER_NODE
N/A
```

treydock:
@ericfranz regarding formatting and headers, with SLURM I think the best option is a very specific `--format` and also passing in `--noheader`.  Parsing header strings to keys just seems problematic when you can build a static map like is currently done.

Not all schedulers and batch systems are as configurable as SLURM when it comes to command output.

Making the adapter version aware I think is best approach.  I've run into cases where major SLURM upgrades not only change CLI formats but also available statuses for jobs so the static mapping would have to be adjusted based on SLURM major version.

ericfranz:
Good question. It might be a parsing error. Looking at the stack trace from https://github.com/OSC/ood-activejobs/issues/169#issuecomment-442702524 I actually don't understand why it breaks.

The stack trace claims it is breaking at https://github.com/OSC/ood_core/blob/9aae58f56f68f4a3a609e0acf1f8b771c31d89b8/lib/ood_core/job/adapters/slurm.rb#L484

Which is calling https://github.com/OSC/ood_core/blob/9aae58f56f68f4a3a609e0acf1f8b771c31d89b8/lib/ood_core/job/adapters/slurm.rb#L425-L430

So it would seem that the wrong value is being assigned to `time` and `:time_used` and then the code that parses that is crashing because it unfortunately assumes if time is not nil it will be of a particular format. Perhaps the output of the value for %b is now including "\x1F" which is the delimiter currently chosen between the output field values:

https://github.com/OSC/ood_core/blob/9aae58f56f68f4a3a609e0acf1f8b771c31d89b8/lib/ood_core/job/adapters/slurm.rb#L85-L96

ericfranz:
For the duration_in_seconds it seems like a regex could improve it (so if a bad value is passed at least it doesn't crash, it might return 0). This appears to cause the tests to pass:

```ruby
# Convert duration to seconds
# 
# valid formats include:
#
#   3-00:00:00 # days hours minutes and seconds
#   22:15:35 # hours minutes and seconds
#   22:30 # minutes and seconds
#   nil
#
def duration_in_seconds(time)
  if time =~ /((?<day>\d+)-)?((?<hours>\d+):)?(?<minutes>\d+):(?<seconds>\d+)/
    ($~['day'].to_i)*24*3600 + ($~['hours'].to_i)*3600 + ($~['minutes'].to_i)*60 + ($~['seconds'].to_i)
  else
    0
  end
end
```

ericfranz:
That is, as a replacement for the existing `duration_in_seconds` implementation. It doesn't address the fact that the parsing might be off.

MorganRodgers:
I just realized what the problem is: it's Slurm. When you pass `%b` to `squeue` it recognizes it, and then doesn't output a delimiter after the output for `%b`. To duplicate submit a job (`--hold` is useful here), and run `squeue -o "%b DELIM %a DELIM %D"`. You'll note that the output for that command will look something like:

```bash
TRES_PER_NODEACCOUNT DELIM NODES
N/Aosc001 DELIM 1
```

Note that there's only 1 DELIM so the `%b` and `%a` fields are effectively concatenated. I'm going to hunt down how to let the Slurm devs know (they don't seem to use Github issues).

Eric points out that we can stupid-patch this by shifting the position of `%b` to the end of the format code.

treydock:
@MorganRodgers SLURM uses bugzilla they host but likely won't respond to bug report unless you are a paying customer.  Everytime I had issues with code that I was capable of fixing they were pretty receptive to github pull requests.  Be glad to help patch SLURM as I've done it in the past.

MorganRodgers:
Reported on their bug tracker: https://bugs.schedmd.com/show_bug.cgi?id=6141

treydock:
Undid my change on 18.08.3 and issue returns:

```
[root@head slurm-18.08.3]# squeue -o "%b DELIM %a DELIM %D"
TRES_PER_NODEACCOUNT DELIM NODES
N/A(null) DELIM 1
```
Looks like suffix printing is missing from current master branch on SLURM github source so likely still an issue.

treydock:
Looks like we have to attach patches to the bugzilla issue you opened.  Also looks like future versions of SLURM will have the handy --parsable flag which I know already exists for `sacct`.

https://github.com/SchedMD/slurm/pull/170

MorganRodgers:
@treydock I don't know if they're going to be difficult about things, but their contributing says that to submit a patch you need to open a new issue with a "severity" of "C - Contribution".

PR #106, opened 2018-11-20
Add An Optional Live System Test For Job Adapters

MorganRodgers:
This branch adds the file [live_system_spec.rb](https://github.com/OSC/ood_core/blob/livesystemtests/spec/job/live_system_spec.rb) which contains a conditional set of tests. If the environment variable `LIVE_CLUSTER_CONFIG` is specified it will attempt to load a cluster config YAML file from that path, and then proceed to check that it can successfully call `submit` on a script that starts off in a held state. Using the `$id` returned by `submit` it will then test `info`, `status`, `release`, and `delete`. Each call to a method on the adapter is in a separate test, so `info` failing will not prevent `delete` from being attempted (which currently happens with Slurm 18). This behavior is useful as it increases the chances of the live job being successfully cleaned up. If `LIVE_CLUSTER_CONFIG` is not defined then the tests will be silently skipped.

Even if the live job is not deleted the user is made aware of the job name so that they can delete it if they want to, and the job is trivial producing no output, and running only for 100 seconds.

Example usage and output:

```
[vagrant@ood ood_core]$ LIVE_CLUSTER_CONFIG=./cluster.yaml scl enable rh-ruby24 -- bundle exec rspec ; squeue
.......................................................................................................................................................................................................................................................................................................................................................................................................................................................
Running tests on live system with job_name TEST_XXDSXAPO
.F...
Live system tests complete. Continuing with other tests:
.....................................................................................................................................

Failures:

  1) a live system can get info
     Failure/Error: time.split(':').map { |v| v.to_i }.inject(0) { |total, v| total * 60 + v }
     
     NoMethodError:
       undefined method `split' for nil:NilClass
     # ./lib/ood_core/job/adapters/slurm.rb:429:in `duration_in_seconds'
     # ./lib/ood_core/job/adapters/slurm.rb:484:in `parse_job_info'
     # ./lib/ood_core/job/adapters/slurm.rb:334:in `block in info'
     # ./lib/ood_core/job/adapters/slurm.rb:333:in `map'
     # ./lib/ood_core/job/adapters/slurm.rb:333:in `info'
     # ./spec/job/live_system_spec.rb:43:in `block (2 levels) in <top (required)>'

Finished in 0.48256 seconds (files took 1.83 seconds to load)
577 examples, 1 failure

Failed examples:

rspec ./spec/job/live_system_spec.rb:41 # a live system can get info

             JOBID PARTITION     NAME     USER ST       TIME  NODES NODELIST(REASON)
[vagrant@ood ood_core]$
```


spec/job/live_system_spec.rb
```diff
@@ -0,0 +1,64 @@
+require 'spec_helper'
+
+describe 'a live system', :if => (! ENV['LIVE_CLUSTER_CONFIG'].nil?) do
```
ericfranz:
```diff
-describe 'a live system', :if => (! ENV['LIVE_CLUSTER_CONFIG'].nil?) do
+describe 'a live system', :if => ENV['LIVE_CLUSTER_CONFIG'] do
or
+describe 'a live system', :unless => ENV['LIVE_CLUSTER_CONFIG'].nil? do
```

ericfranz: You should also add `:order => :defined` so that you guarantee these tests run in order.

https://relishapp.com/rspec/rspec-core/v/3-8/docs/configuration/overriding-global-ordering#running-a-specific-example-group-in-order

spec/job/live_system_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+      #!/bin/bash
+      #JOB_HEADERS_HERE
+      sleep 100
+      exit 0
+    HERESCRIPT
+
+    @script = OodCore::Job::Script.new(
+      # generate a random job name
+      # e.g. TEST_SYQECQXW
+      job_name: 'TEST_' + (0...8).map { (65 + rand(26)).chr }.join,
```
ericfranz: In Ruby 2.5 we get SecureRandom.alphanumeric. But until then :-).

spec/job/live_system_spec.rb
MorganRodgers: Updated.

ericfranz:
Looks good to me. Please resolve CHANGELOG conflicts and merge.

Issue #107, opened 2018-11-26
Merge pbs-ruby into ood_core


ericfranz:
This introduces the `ffi` dependency to ood_core and remove the pbs gem dependency.

Could under `job/adapters/torque/` directory which probably makes more sense and then do this:

| pbs-ruby  |  ood_core |
| ------------- | ------------- |
| `PBS`  | `OodCore::Job::Adapters::Torque`  |
| `PBS::Batch`  | `OodCore::Job::Adapters::Torque::Batch`  |
| `PBS::Torque`  | `OodCore::Job::Adapters::Torque::FFI`  |

* note that if you rename the `PBS::Torque` module to `FFI` then references to the main `FFI` class would need the `::` prefix i.e. `extend ::FFI::Library` and `class Attrl < ::FFI::Struct` and `::FFI::Pointer::NULL`. But that would probably be a better name since that module is the list of functions that provide the FFI interface. Also worth looking at other projects that use FFI for better naming inspiration. Other name could be `OodCore::Job::Adapters::Torque::CLib` or `OodCore::Job::Adapters::Torque::C` since it is the C library interface; the fact that we use FFI is an implementation detail (and arguably we could switch to Fiddle later anyways if there was a use case).

MorganRodgers:
I like the idea of bringing the PBS adapter into `ood_core`. We are only going to use and support it in the context of OnDemand, so I agree that it doesn't make sense as a stand alone dependency.

MorganRodgers:
Todo:

* [x] Initial git merge of projects
* [ ] Fix names
* [ ] Fix tests

PR #108, opened 2018-11-27
Merge pbs-ruby into ood_core

MorganRodgers:
Addresses #107.

I performed the merge by adding a remote and pulling, so the entire history of the pbs-ruby master branch will be preserved.

That also means that the 'changes' in this PR are really only visible in 0dd70d5 to HEAD.


ericfranz:
1. The module name over batch, attributes, and error are wrong. The module should be `OodCore::Job::Adapters::Torque` which corresponds to directory structure `ood_core/job/adapters/torque/`. And actually this is a class, so instead of `module OodCore::Job::Adapters::Torque` at the top it would be `class OodCore::Job::Adapters::Torque` because that class is first defined 
2. torque_ffi.rb should be renamed to ffi.rb to match the ruby naming convention since the classes name is FFI. i.e.:

   ```ruby
   class OodCore::Job::Adapters::Torque
     module FFI
       ...
     end
   end
   ```

3. After these changes in the class `OodCore::Job::Adapters::Torque::Batch#connect` you may be able to get by, instead of doing `OodCore::Job::Adapters::Torque::FFI.lib` just doing `FFI.lib` as the `FFI` would be `OodCore::Job::Adapters::Torque::` not `::FFI`.
4. There is a stray comment above the torque/batch_spec.rb that can be removed

ericfranz:
I can see how when reading the code that reading `FFI.lib` it might confuse you that this is `FFI` in the Torque module. So if you want to keep the class `OodCore::Job::Adapters::Torque` and add `OodCore::Job::Adapters::Torque::TorqueFFI` for the sole reason that in the Batch class you can do `TorqueFFI.lib` and `TorqueFFI.pbs_connect`. Actually I thought we discussed before it might be preferable to call this `CLib`. Then you would have the module `OodCore::Job::Adapters::Torque::CLib` and in the `OodCore::Job::Adapters::Torque::Batch` class you could do `CLib.lib` and `CLib.pbs_connect` or `C`? Something that indicates what it is (C library interface) vs how it is implemented (FFI)?

Though FFI does indicate it is a "foreign function". Ugh. Naming things.

ericfranz:
Final observations:

1. torque_ffi.rb batch.rb  whitespace is off - there is an two space indentation for line `class Batch` and for the entirety of the code in `torque_ffi.rb`
2. rename the file torque_ffi.rb to ffi.rb (actual code changes not necessary for that change) and of course update the require statement in torque.rb

ericfranz:
Then assuming its passing the tests just merge it!

Issue #111, opened 2018-12-03
Configure custom paths to the commands used to submit jobs


ericfranz:
You could extend this, possibly to any command used to shell out by an adapter - though it would not work for adapters that use C lib for getting job status for example.

It seems we have several use cases where being able to provide a custom wrapper around `qsub` or `sbatch`, without having to wrap the rest of the commands, would be useful. Currently you can only specify path to bin in the cluster config. The idea is you could do this:

```yaml
  job:
    adapter: "torque"
    host: "owens-batch.ten.osc.edu"
    lib: "/opt/torque/lib64"
    bin: "/opt/torque/bin"
    custom_bin:
      qsub: "/opt/torque-osc-custom/bin/qsub"
    version: "6.0.1"
```

And then when qsub is executed instead of doing `join('/opt/torque/bin', 'qsub')` to build the path to execute `qsub` is used.

Presumably this could be done in a simple way: use a hash to store these custom paths, and then have the default be a block that joins the bin path. So instead of https://github.com/OSC/ood_core/blob/602e077e761ede2cc8bc241560918423330b3db5/lib/ood_core/job/adapters/slurm.rb#L137-L144

We would do something like:

```ruby
  call("sbatch", *args, env: env, stdin: str.to_s).strip.split(";").first
end

private

def call(cmd, *args, env: {}, stdin: "")
  cmd = bin_path(cmd)
  # ... rest of method
end

def bin_path(cmd)
  custom_bin.fetch(cmd.to_sym) { bin.join(cmd.to_s).to_s }
end
```

Where `custom_bin` returns the hash form of the custom bin section, and the default value is the preexisting functionality of joining bin with the command.

We would just need to document which adapters used shell commands (and nothing else) to know what could be safely modified.

MorganRodgers:
This breaks for things like SGE(coming soon™) (which can use DRMAA for `#info`) and for Torque which is using `libtorque.so` via the PBS gem. I am curious is Torque like SGE in that there are just huge advantages to using the FFI adapter rather than parsing command line output?

ericfranz:
Both SGE and Torque use `qsub` so we would just have to document, per adapter, what is supported. The adapters could even have a list of legal custom_bin keys and if you try to launch with an illegal one it raises an exception (not implemented or supported or something).

MorganRodgers:
Was merged in #112

PR #112, opened 2018-12-05
Custom bin

MorganRodgers:
Extends the adapters and the cluster config to allow overriding certain scheduler client executables as described in #111.

If we move ahead with this we will also need to update the docs to:

- [x] Document this extension
- [x] Warn users that this is an 'advanced' feature because OOD expects to capture return codes and parses `stdout`; wrappers which alter either of those things from the default risk breaking OOD functionality
    - [ ] Consider adding a well documented example of "how to do it right" for wrappers for `qsub` and `sbatch`?
    - [ ] This would be a good spot to also promote the live system test for adapters #106
- [ ] Make mention that because the Torque adapter uses `libtorque.so` the only overridable client is `qsub`.


lib/ood_core/job/adapters/pbspro.rb
```diff
@@ -44,15 +47,22 @@ class Batch
           # @return [Pathname, nil] path to pbs executables
           attr_reader :exec
```
ericfranz: `exec` is a Kernel method - lets fix this to be `bin` like the other adapters. I guess it is written to mirror the configuration hash which is:

https://osc.github.io/ood-documentation/master/installation/resource-manager/pbspro.html?highlight=pbspro

and contains `exec: "/path/to/pbspro"`. Is there a PBSPro reason why this is the case? Perhaps we should change this by default to be bin, and do something like this in the `build_pbspro` factory method

```
bin = c.fetch(:bin) { c.fetch(:exec, nil) }
```

This will also make sense then so that the naming is similar (bin, custom_bin vs exec, custom_bin). Then of course the `PBSPro::Batch` would be updated so that `attr_reader :exec` is `attr_reader :bin`.

lib/ood_core/job/adapters/pbspro.rb
```diff
@@ -44,15 +47,22 @@ class Batch
...
 
+          # Optional overrides for PBS Pro client executables
+          # @example
+          #  {'qsub' => '/usr/local/bin/qsub'}
+          # @return Hash<String, String>
+          attr_reader :custom_bin
```
ericfranz: I wonder if calling this `custom_bin_paths` would be more readable. When I was just glancing at the code, especially the implementation of `OodCore::Job::Adapters::Helper.bin_path` it wasn't clear to me that `bin` and `custom_bin` were two different types dealing with different directories. Though `custom_bin_paths` isn't that much clearer. Any ideas on a better name? if not we can keep it as is.

lib/ood_core/job/adapters/torque/batch.rb
```diff
@@ -440,15 +447,17 @@ def qsub_submit(script, queue, headers, resources, envvars)
           "PBS_DEFAULT"     => "#{host}",
           "LD_LIBRARY_PATH" => "#{lib}:#{ENV['LD_LIBRARY_PATH']}"
         }
-        cmd = bin.join("qsub").to_s
+        cmd = OodCore::Job::Adapters::Helper.bin_path('qsub', bindir, custom_bin)
```
ericfranz: Did you mean `bin` instead of `bindir` here?

lib/ood_core/job/adapters/torque/batch.rb
```diff
@@ -440,15 +447,17 @@ def qsub_submit(script, queue, headers, resources, envvars)
...
+        p cmd
```
ericfranz: left print statement here

lib/ood_core/job/adapters/torque/batch.rb
```diff
@@ -440,15 +447,17 @@ def qsub_submit(script, queue, headers, resources, envvars)
...
         o, e, s = Open3.capture3(env, cmd, *params)
         raise Error, e unless s.success?
         o.chomp
       end
 
       # Call a forked PBS command for a given host
       def call(cmd, *args, env: {}, stdin: "", chdir: nil)
-        cmd  = bin.join(cmd.to_s).to_s
+        cmd = OodCore::Job::Adapters::Helper.bin_path(cmd, bin, custom_bin)
+        p cmd
```
ericfranz: left print statement here

ericfranz:
I have three thoughts after looking at the code:

1. There seems to be preexisting problems with not correctly handling nil case (though our documentation does ask for the full paths to be specified in the cluster config). Maybe the helper `bin_path` should just return the command if default_bin is nil? or empty? and a custom bin path is not specified. 
2. Naming is difficult. But `custom_bin` as a name (which I know I suggested) doesn't seem ideal - as it is very close to `bin` and yet `bin` refers to the bin directory containing the binaries, whereas `custom_bin` refers to the hash of the paths to custom binaries themselves. But maybe just good comments are good enough.
3. The helper is nice in that you have the simple `bin_path(command, default_bin, custom_bin)`. And what I'm about to suggest might introduce slightly more duplication (unless you use metaprogramming). But it might be clearer code, to instead of having these methods pass a string argument to "command", and then have the `call` method be responsible for building the full path to the command, to instead add methods to get the full command directly.

For example the PBSPro adapter. We could define:

```ruby
class OodCore::Job::PBSPro::Batch

  # if we are following the convention of the other batch objects, the bin method
  # returns the Pathname to the bin directory, or nil if not set
  def bin
    (!!exec) ? exec.join("bin") : nil
  end

  %w(qsub qstat qdel qhold qselect).each do |cmd|
    define_method("#{cmd}_path") do
      OodCore::Job::Adapters::Helper.bin_path(cmd, bin, custom_bin)
    end
  end
```

Then you do

```diff
  def delete_job(id)
-   call("qdel", id.to_s)
+   call(qdel_path, id.to_s)
  end

    def submit_string(str, args: [], chdir: nil)
-     call("qsub", *args, stdin: str.to_s, chdir: chdir).strip
+     call(qsub_path, *args, stdin: str.to_s, chdir: chdir).strip
    end
```

Finally, I just noticed this:

```
irb(main):007:0> Pathname.new('').join(Pathname.new('qsub')).to_s
=> "qsub"
irb(main):009:0> nil.to_s
=> ""
```

This means that `OodCore::Job::Adapters::Helper.bin_path` could be updated to handle the nil edge case and simplified by doing:

```
Pathname.new(custom_bin.fetch(bin).to_s).join(cmd)
```

ericfranz:
Finally a good comment on `OodCore::Job::Adapters::Helper.bin_path` might be useful.

Only other thought. I used to always suggest creating objects that you would call new on to use, since it is often easier to work with (easier to extend, easier to mock, etc.) than have class methods. So in the same way It might be the case that, just like https://github.com/OSC/ood_core/blob/master/lib/ood_core/job/adapters/lsf/helper.rb and https://github.com/OSC/ood_core/blob/sge/lib/ood_core/job/adapters/sge/helper.rb, `OodCore::Job::Adapters::Helper` is a class that we call new on. So a direct call would change from `OodCore::Job::Adapters::Helper.bin_path` to `OodCore::Job::Adapters::Helper.new.bin_path` or more simply we could add a getter on the batch objects:

```ruby
def helper
  @helper ||= OodCore::Job::Adapters::Helper.new
end

# then use it like so
helper.bin_path
```

Of course we would resolve the name conflict between the generic helper and the adapter specific helpers.

MorganRodgers:
How do you feel about replace `custom_bin` with `bin_overrides`?

ericfranz:
All of this is thoughts for refactoring that could definitely wait till later. Regardless, above `OodCore::JobAdapters::Helper` file a comment at the top that includes `# @api private` so we are a saying we can change in the future and we don't consider it part of the "public interface".

lib/ood_core/job/adapters/torque/batch.rb
MorganRodgers: Yes, yes I did mean that.

lib/ood_core/job/adapters/torque/batch.rb
MorganRodgers: Deleted.

lib/ood_core/job/adapters/torque/batch.rb
MorganRodgers: also deleted.

MorganRodgers:
I elected not to change much in `PBSPro`. Your solution for simplifying `OodCore::Job::Adapters::Helper.bin_path` behaves differently than the original (and not just because of the return type). The original allows for e.g. `/opt/slurm/bin/squeue` to be replaced by `/usr/local/bin/something_not_named_squeue`. The proposed alternative would require sites to name their wrappers after the command that they are replacing, which they may not want to do.

ericfranz:
Oops that was a mistake on my part - I didn't intend on the behavior to differ.

Then I would fix the implementation like this:

```ruby
def self.bin_path(cmd, std_bin, bin_overrides)
  bin_overrides.fetch(cmd.to_s) { Pathname.new(std_bin.to_s).join(cmd.to_s).to_s }
end
```

That ensures if std_bin is nil, and bin_overrides is {} you get back "squeue" instead of a crash, which is better functionality.

```
irb(main):029:0> def bin_path(cmd, std_bin, bin_overrides)
irb(main):030:1>   bin_overrides.fetch(cmd.to_s) { std_bin.join(cmd.to_s).to_s }
irb(main):031:1> end
=> :bin_path
irb(main):032:0> bin_path("squeue", nil, {})
Traceback (most recent call last):
        5: from /Users/efranz/.asdf/installs/ruby/2.5.0/bin/irb:11:in `<main>'
        4: from (irb):32
        3: from (irb):30:in `bin_path'
        2: from (irb):30:in `fetch'
        1: from (irb):30:in `block in bin_path'
NoMethodError (undefined method `join' for nil:NilClass)
irb(main):017:0> def bin_path(cmd, std_bin, bin_overrides)
irb(main):018:1>   bin_overrides.fetch(cmd.to_s) { Pathname.new(std_bin.to_s).join(cmd.to_s).to_s }
irb(main):019:1> end
=> :bin_path
irb(main):026:0> bin_path("squeue", nil, {})
=> "squeue"
irb(main):027:0> bin_path("squeue", "/opt/slurm/bin/squeue", {})
=> "/opt/slurm/bin/squeue/squeue"
irb(main):028:0> bin_path("squeue", "/opt/slurm/bin/squeue", custom_bin)
=> "/usr/local/bin/something_not_named_squeue"
```

ericfranz:
Final minor gripe: is `default_bin` or `bin_default` preferable to `std_bin`? Since the comment it self says "the default place to find that command". It also flows nicer (bin_default, bin_overrides) and then matches with the Batch objects (bin, bin_overrides). I guess thats better (think to much about naming things and you second guess everything).

MorganRodgers:
Naming things is easy: just let your fingers choose and never look back. ;)

ericfranz:
One final thing - the code in call itself is not exercised by tests. So you can actually do this:

```diff
efranz@webtest04:~/dev/ood_core (custom_bin)$ git diff lib/ood_core/job/adapters/pbspro.rb
diff --git a/lib/ood_core/job/adapters/pbspro.rb b/lib/ood_core/job/adapters/pbspro.rb
index e7d7c2c..29bf71c 100644
--- a/lib/ood_core/job/adapters/pbspro.rb
+++ b/lib/ood_core/job/adapters/pbspro.rb
@@ -158,7 +158,7 @@ module OodCore
             def call(cmd, *args, env: {}, stdin: "", chdir: nil)
               cmd = cmd.to_s
               bindir = (!!pbs_exec) ? pbs_exec.join("bin").to_s : ''
-              cmd = OodCore::Job::Adapters::Helper.bin_path(cmd, bindir, bin_overrides)
+              cmd = OodCore::Job::Adapters::Helper.bin_path(cmd, bindirr, bin_overrides)
               args = args.map(&:to_s)
               env = env.to_h.each_with_object({}) { |(k, v), h| h[k.to_s] = v.to_s }
               env["PBS_DEFAULT"] = host.to_s if host
```

And run the tests and they will still pass.

The fix is to mock Open3 for this particular test. I submitted a separate branch/pr as an example. We should probably add similar tests to each of the adapters.

ericfranz:
See https://github.com/OSC/ood_core/pull/113

The problem with this code is the test is difficult to read IMHO. I don't know if I'm just not used to rspec. But the only thing under test is the **second argument**, and everything else on the second line is just mocking the call.

Note that I will admit we didn't have tests like this before for other code, which is one of the reasons why I detest that we have so much complexity in a method like `call` which requires mocking in order to test (and as a result, often doesn't have tests).

ericfranz:
PR is an idea, if you have better, by all means.

PR #118, opened 2018-12-26
Set SGE_ROOT even when libdrmaa is not used

ericfranz:
It ends up that at least in some versions of SGE, SGE_ROOT needs
to be defined in the environment, even if not using libdrmaa.

So for now:

1. Always set SGE_ROOT to the @sge_root string value or ENV['SGE_ROOT'] or a default value.
2. Require explicit path to libdrmaa to be set in order to use libdrmaa.


ericfranz:
This was tested with libdrmaa and without on the test Vagrant image.

PR #126, opened 2019-02-08
PBS job array support

MorganRodgers:
Fixes #126 

Job array support for PBSPro


MorganRodgers:
Note: https://github.com/MorganRodgers/pbspro-workbench for testing things.

ericfranz:
When doing a qstat to get all the jobs, what about optionally omitting `-t`? Would a parent id of a job array job still appear, but omit all of the job tasks?

See https://github.com/OSC/ood-activejobs/issues/131. Perhaps this is something to re-address in 1.6, the idea of not showing job array tasks in active jobs list view and only in the details view showing the tasks and each of their statuses.

MorganRodgers:
PBSPro's `qstat` will list the parent job with out `-t` being used:

```sh
[vagrant@head bin]$ qstat
Job id            Name             User              Time Use S Queue
----------------  ---------------- ----------------  -------- - -----
3[].head          STDIN            ood                      0 B workq           
[vagrant@head bin]$ qstat -t
Job id            Name             User              Time Use S Queue
----------------  ---------------- ----------------  -------- - -----
3[].head          STDIN            ood                      0 B workq           
3[1].head         STDIN            ood               00:00:00 R workq           
3[2].head         STDIN            ood                      0 Q workq           
...
```

To add the option we would want to add it to the cluster config. Something like:

```yaml
---
v2:
  metadata:
    title: "PBSPro Cluster"
  login:
    host: "ood"
  job:
    adapter: "pbspro"
    host: "head"
    exec: "/opt/pbs/"
    # vvv New Stuff vvv #
    options:  # or settings?
      # Called by ActiveJobs
      info_all:
        # Optionally pass -t to show job array tasks; defaults true
        show_job_array_tasks: false
```

lib/ood_core/job/info.rb
```diff
@@ -113,6 +113,21 @@ def initialize(id:, status:, allocated_nodes: [], submit_host: nil,
         @native          = native
       end
 
+      # Create a new Info for a child task
+      # @return [Info] merging the parent and the child task
+      def new_for_child_task(task)
```
ericfranz: I'd rename this to `build_child_task`. 

Inspiration is the method naming conventions in Rails for Active Record collection build methods: https://apidock.com/rails/ActiveRecord/Associations/CollectionProxy/build and https://guides.rubyonrails.org/association_basics.html#methods-added-by-belongs-to-build-association-attributes

ericfranz: Oh except it isn't a child task that is being built. The equivalent "build" name for new_for_child_task would be something like `build_child_info`

MorganRodgers:
Fixes #132.

lib/ood_core/job/adapters/pbspro.rb
```diff
@@ -85,9 +85,9 @@ def initialize(host: nil, pbs_exec: nil, bin_overrides: {})
           # @param id [#to_s] the id of the job
           # @raise [Error] if `qstat` command exited unsuccessfully
           # @return [Array<Hash>] list of details for jobs
-          def get_jobs(id: "")
-            args = ["-f"]   # display all information
-            args += ["-t"]  # list subjobs
+          def get_jobs(id: "", user: "")
+            args = ["-f", "-t"]   # display all information
+            args += ["-u", user] unless user.empty?
```
ericfranz: The new user option doesn't appear to be used anywhere... is there a plan to?

MorganRodgers: It's a remnant of a plan to re-write `info_where_owner` that I abandoned because it would have changed behavior. I'll remove it.

lib/ood_core/job/adapters/pbspro.rb
```diff
@@ -434,6 +453,21 @@ def parse_job_info(v)
               native: v
             )
           end
+
+          # Combine the array parent with the states of its children
+          def process_job_array(jobs)
+            parent = jobs.select { |j| /\[\]/ =~ j.id }.first.to_h
```
ericfranz: I guess its a fair assumption, but this code assumes the array of jobs has at least one `Info` object with the id in the form of "parent[]". Crash occurs if thats not the case (if first returns nil)

MorganRodgers: This method is only called once and is protected by an if/elsif: https://github.com/OSC/ood_core/blob/pbs-array/lib/ood_core/job/adapters/pbspro.rb#L336-L342

PR #130, opened 2019-02-28
LSF job arrays

### LINKED ISSUE #131: Add LSF Job Array Support

MorganRodgers:
Fixes #131 

Update LSF to handle job arrays.

The code is fit to be reviewed. I still need to write a few regular tests to cover cases that were only caught by the live system tests. Once that is done I will remove the WIP label.


ericfranz:
What happens if you submit a job array with no job name?

MorganRodgers:
> What happens if you submit a job array with no job name?

That is a test case I should write up. The adapter will submit the job, and the scheduler will recognize the job array spec and schedule it as such. When running `adapter.info(id)` the `Tasks` will have incorrect task ids.

MorganRodgers:
@ericfranz I addressed your question about submitting a job array with no name.

lib/ood_core/job/adapters/lsf.rb
```diff
@@ -196,8 +193,11 @@ def info_for_batch_hash(v)
             dispatch_time = helper.parse_past_time(v[:start_time], ignore_errors: true)
             finish_time = helper.parse_past_time(v[:finish_time], ignore_errors: true)
 
+            # Detect job array index from name
+            array_index = /^[^\[]*(\[[0-9]+\])$/.match(v[:name])
```
ericfranz: From what I gather, the assumptions here is LSF array jobs will have the id suffix the Job Name i.e. myjob[1]. So for a job id 1234, to get "1234[1]" you need to capture this index.

If so you do not need to match the entire string. This will capture the same thing:

```ruby
array_index = /(\[\d+\])$/.match(v[:name])
```

That regex should be faster since it should scan the end of the string instead of the entire string.

MorganRodgers: An additional bonus is that your expression is simpler.

ericfranz: Another approach is "even faster" but we probably wouldn't see it without millions of jobs:

```ruby
/^[^\[]*(\[[0-9]+\])$/ =~ v[:name]
Info.new(
  id: "#{v[:id]}#{$1.to_s}"
```

Its faster cause `$1` already exists so that is just assigned `nil` or `[1]` etc. and then since you have nil or your string you can just call to_s to coerce...

MorganRodgers: I appreciate the elegance of that, but I don't think that it is particularly clear code, and if we wouldn't see a large performance advantage then it isn't worth the loss of clarity.

ericfranz: Its 15-20 ms if there are 20000 jobs.

```
irb(main):068:0> Benchmark.measure { 20000.times { /(\[\d+\])$/ =~ "myjob[1]"; id = "1234#{$1}"; } }.real*1000
=> 10.776999988593161
irb(main):069:0> Benchmark.measure { 20000.times { /(\[\d+\])$/ =~ "myjob[1]"; id = "1234#{$1}"; } }.real*1000
=> 10.088999988511205
irb(main):070:0> Benchmark.measure { 20000.times { /(\[\d+\])$/ =~ "myjob[1]"; id = "1234#{$1}"; } }.real*1000
=> 10.46899997163564
irb(main):071:0> Benchmark.measure { 20000.times { /(\[\d+\])$/ =~ "myjob[1]"; id = "1234#{$1}"; } }.real*1000
=> 10.78599999891594
irb(main):072:0> Benchmark.measure { 20000.times { ary_idx = /(\[\d+\])$/.match("myjob[1]"); id = "1234#{ary_idx}"; } }.real*1000
=> 28.82300002966076
irb(main):073:0> Benchmark.measure { 20000.times { ary_idx = /(\[\d+\])$/.match("myjob[1]"); id = "1234#{ary_idx}"; } }.real*1000
=> 28.51200004806742
irb(main):074:0> Benchmark.measure { 20000.times { ary_idx = /(\[\d+\])$/.match("myjob[1]"); id = "1234#{ary_idx}"; } }.real*1000
=> 28.21700001368299
irb(main):075:0> Benchmark.measure { 20000.times { ary_idx = /(\[\d+\])$/.match("myjob[1]"); id = "1234#{ary_idx}"; } }.real*1000
=> 28.084999998100102
```

So the gain is small but I understated it by saying we wouldn't see it "without millions of jobs". There are bigger wins to be had elsewhere though, as the cost here is due to the creation of the separate match object many times (instead of reusing a preexisting var `$1`). We could revisit as a broad sweep of "optimizing regex use in adapters" if its worth it. At that point, we might consider taking a different strategy though.

lib/ood_core/job/adapters/lsf.rb
```diff
@@ -90,15 +90,18 @@ def submit(script, after: [], afterok: [], afternotok: [], afterany: [])
...
+          when 0
             Info.new(
               id: id,
               status: :completed
             )
+          when 1
+            info_ary.first
+          else  # many
+            handle_job_array(info_ary, id)
           end
```
ericfranz: It might be more readable to just handle the edge cases at the top of handle_job_array:

```ruby
def info(id)
  info_ary = batch.get_job(id: id).map{|v| info_for_batch_hash(v)}
  handle_job_array(info_ary, id)
rescue Batch::Error => e
  # ...
end

def handle_job_array(info_ary, id)
  return Info.new(id: id, status: :completed) if info_ary.nil? || info_ary.empty?
  return info_ary.first if info_ary.size == 1

  parent_task_hash = build_proxy_parent(info_ary.first, id)
  #...
```

PR #133, opened 2019-03-06
SGE defaults to passing the -cwd flag

MorganRodgers:
By setting `-cwd` we ensure that stdout/stderr are written to the correct place. With GE a script's content does not override command line options to `qsub`. The GE helper attempts to work around the `qsub` limitations by detecting `-wd` and `-cwd` directives in the first 1024 characters and first 20 lines of a script, thus allowing us to be selective about whether we want to force the `-cwd` flag to be set or not.

Tests were written for the helper method. Note as well that `-cwd` became a default option for the `#submit` tests, necessitating its inclusion.

### DIFF
--- a/lib/ood_core/job/adapters/sge/helper.rb
+++ b/lib/ood_core/job/adapters/sge/helper.rb
@@ -20,7 +20,12 @@ class OodCore::Job::Adapters::Sge::Helper
     args += ['-h'] if script.submit_as_hold
     args += ['-r', 'yes'] if script.rerunnable
     script.job_environment.each_pair {|k, v| args += ['-v', "#{k.to_s}=#{v.to_s}"]} unless script.job_environment.nil?
-    args += ['-wd', script.workdir] unless script.workdir.nil?
+
+    if script.workdir
+      args += ['-wd', script.workdir]
+    elsif ! script_contains_wd_directive?(script.content)
+      args += ['-cwd']
+    end
 
     on_event_email = []
     on_event_email << 'b' if script.email_on_started  # beginning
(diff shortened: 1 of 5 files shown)


lib/ood_core/job/adapters/sge/helper.rb
```diff
@@ -47,6 +52,41 @@ def batch_submit_args(script, after: [], afterok: [], afternotok: [], afterany:
...
+  #
+  # @param      content  The script content
+  # 
+  # Examples:
+  #     #$-wd /home/ood/ondemand  # should match
+  #     #$  -wd /home/ood/ondemand  # should match
+  #     #$  -cwd /home/ood/ondemand  # should match
+  #     #$ -j yes -wd /home/ood/ondemand  # should match
+  #     #$ -j yes -o this-wd /home/ood/ondemand  # should NOT match
+  #       #$ -t 1-10:5 -wd /home/ood/ondemand  # should NOT match
```
ericfranz: this doesn't match because there is whitespace before `#$`?

MorganRodgers: Correct.

lib/ood_core/job/adapters/sge/helper.rb
```diff
@@ -47,6 +52,41 @@ def batch_submit_args(script, after: [], afterok: [], afternotok: [], afterany:
...
+  def script_contains_wd_directive?(content)
+    content.slice(
+      # Only search within the script's first 1024 characters in case the user is
+      # putting non-line delimited data into their scripts.
+      0, 1024
+    ).split(
+      "\n"
+    ).slice(
+      # Only scan in the first 20 rows
+      0, 20
```
ericfranz: Is there any reason to do this now that we do the first 1024 chars?

MorganRodgers: I suppose not.

spec/job/adapters/sge/helper_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+describe OodCore::Job::Adapters::Sge::Helper do
+  subject(:helper) {described_class.new}
+
+  describe "#script_contains_wd_directive?" do
+    let(:match_w_no_space) {'#$-wd /home/ood/ondemand'}
+    let(:match_leading_space) {'#$  -wd /home/ood/ondemand'}
+    let(:match_cwd) {'#$  -cwd /home/ood/ondemand'}
+    let(:match_multiple_directives) {'#$  -wd /home/ood/ondemand'}
+
+    let(:should_not_match_embedded_wd) {' #$ -j yes -o this-wd /home/ood/ondemand'}
```
ericfranz: for this test, the whitespace before `#$` should not exist

Issue #135, opened 2019-03-07
Remove or disable "comment" from the fields for squeue?


ericfranz:
At TAMU and possibly other sites, "%k" will output multiple lines.

This causes problems parsing the results since it is assumed that 100% of the data will end up on one line.

Instead of assuming a newline separator for each job, use an explicit record separator i.e. replace https://github.com/OSC/ood_core/blob/29cfbb243fff817741159ecb35cd97b2b2e49ec7/lib/ood_core/job/adapters/slurm.rb#L95-L105

with something like

```ruby
def get_jobs(id: "", filters: [])
  record_separator = "\x1E"
  unit_seperator = "\x1F"

  # ...

  args += ["-o", "#{options.values.join(unit_seperator)}"]
  args += ["-j", id.to_s] unless id.to_s.empty?
  lines = call("squeue", *args).split(record_separator).map(&:strip).compact
 
  # ...
end
```

Note we want to improve this anyway to take a streaming approach, instead of loading all of the lines at once.

MorganRodgers:
Fixed in #138.

PR #136, opened 2019-03-08
[WIP] Use record separators instead of newlines for squeue parsing

ericfranz:
See https://en.m.wikipedia.org/wiki/Delimiter#Conventions

If we used record separators instead of newlines then we could address edge cases like https://github.com/OSC/ood_core/issues/135 where %k output has newlines.


ericfranz:
Addressed in https://github.com/OSC/ood_core/pull/138

Issue #134, opened 2019-03-07
If submit time is not available Slurm adapter crashes on squeue parsing


ericfranz:
submit_time is nil then this will crash https://github.com/OSC/ood_core/blob/29cfbb243fff817741159ecb35cd97b2b2e49ec7/lib/ood_core/job/adapters/slurm.rb#L494

this was experienced at a site where %k was producing multiple lines, introducing parsing issues

ericfranz:
Easy way to test this is if we have a function to convert a slurm hash to an Info hash.

MorganRodgers:
Fixed in #138, but without a separate test.

Issue #121, opened 2019-01-23
Slurm adapter should implement info_where_owner for speed improvements


ericfranz:

MorganRodgers:
From the `squeue` manpage:

```man
       -u <user_list>, --user=<user_list>
              Request jobs or job steps from a comma separated list of users.  The list can consist of user names
              or user id numbers.  Performance of the command can be measurably improved for systems  with  large
              numbers of jobs when a single user is specified.
```

MorganRodgers:
Added in #139.

PR #138, opened 2019-03-18
[WIP] Slurm: use \x1E record separator and refactor adapter

ericfranz:


ericfranz:
filters, fields, attrs, options => four names for basically the same things :-(. Not sure a better way to have consistent naming. The argument for ood_core is attrs so maybe we should stick with that (is fields better)? squeue though calls them options and fields.

>-o <output_format>, --format=<output_format>
>
> Specify the information to be displayed, its size and position (right or left justified). Also see the -O <output_format>, --Format=<output_format> option described below (which supports less flexibility in formatting, but supports access to all fields). The default formats with various options are 

* https://slurm.schedmd.com/squeue.html
* Well I guess after a closer reading, it calls "options" for the format types (default, -l, -s) versus "fields" referring to each column i.e. "%M Time used by the job or job step... For job steps this field shows the...." and "%all Print all fields available..."

ericfranz:
If fields were more appropriate it would appear that fields should be used instead of attrs for the info_all argument i.e.:

`info_all(fields: [:job_id, :status, :job_name, :queue_name])` versus `info_all(attrs: [:job_id, :status, :job_name, :queue_name])`

ericfranz:
TODO:

- https://github.com/OSC/ood_core/issues/121
- https://github.com/OSC/ood_core/issues/134
- streaming `info_all_each`

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -89,22 +95,66 @@ def initialize(cluster: nil, bin: nil, conf: nil, bin_overrides: {})
...
-            options = filters.empty? ? fields : fields.slice(*filters)
-            args  = ["--all", "--states=all", "--noconvert"]
-            args += ["-o", "#{options.values.join(delim)}"]
-            args += ["-j", id.to_s] unless id.to_s.empty?
-            lines = call("squeue", *args).split("\n").map(&:strip)
+          def get_jobs(id: "", attrs: nil, record_separator: RECORD_SEPARATOR, unit_separator: UNIT_SEPARATOR)
+            fields = squeue_fields(attrs)
+            args = squeue_args(id: id, options: fields.values, record_separator: record_separator, unit_separator: unit_separator)
+
+            #TODO: switch mock of Open3 to be the squeue mock script
```
MorganRodgers: Note that the "switch mock of Open3 to be the squeue mock script" is not in your TODO list.

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -89,22 +95,66 @@ def initialize(cluster: nil, bin: nil, conf: nil, bin_overrides: {})
...
+            # then you can use that for performance metrics
+            StringIO.open(call("squeue", *args)) do |output|
+              # if cluster, skip 2 lines, else skip 1 line
+              (cluster ? 2 : 1).times { output.gets(RECORD_SEPARATOR) }
```
MorganRodgers: I think you meant this to be the method parameter `record_separator` instead of the constant `RECORD_SEPARATOR`.

Also, why parameterize for that? Do you anticipate that we may need to vary the separator characters? If so then why use constants?

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -147,6 +197,62 @@ def submit_string(str, args: [], env: {})
...
+              submit_time: "%V",
+              wckey: "%w",
+              licenses: "%W",
+              excluded_nodes: "%x",
+              core_specialization: "%X",
+              nice: "%y",
+              scheduled_nodes: "%Y",
+              sockets_cores_threads: "%z",
+              work_dir: "%Z",
+              gres: "%b",  # must come at the end to fix a bug with Slurm 18
```
MorganRodgers: Tangential to this review: an outcome of the last ops meeting was the revelation that `gres`/`tres` is how Slurm allows GPUs to be a Trackable Resource. We may end up caring more about this field in the future.

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -369,7 +440,7 @@ def status(id)
           id = id.to_s
           jobs = @slurm.get_jobs(
             id: id,
-            filters: [:job_id, :array_job_task_id, :state_compact]
+            attrs: [:job_id, :array_job_task_id, :state_compact]
```
MorganRodgers: This is a good decision. I prefer that we use `filters` for when we are attempting to filter which records are returned rather than which record attributes are returned.

lib/ood_core/job/adapters/slurm.rb
ericfranz: Added. The idea is to add a script to a spec/fixtures/ directory that outputs https://github.com/OSC/ood_core/blob/4ee4b03e9997e5381016ee2742e053fe29d99fda/spec/job/adapters/slurm_spec.rb#L260-L262 so we can use bin_overrides instead of doing

```
allow(Open3).to receive(:capture3).and_return([squeue_output, "", double("success?" => true)])
```

lib/ood_core/job/adapters/slurm.rb
ericfranz: Premature optimization. Thanks. Removed!

lib/ood_core/job/adapters/slurm.rb
ericfranz: Oh boy!

ericfranz:
@MorganRodgers none of the adapters support attrs yet. So here are two questions:

1. The intention (as defined and currently used in Active Jobs) is if we specify attrs: [] or https://github.com/OSC/ood-activejobs/blob/master/app/models/jobs_json_request_handler.rb#L5 we will get back at least "required args" such as the job id and the status. Part of me feels like this is bad - we should just allow specifying exactly what you want (even though we only support this on info_all and why would you do info_all without specifying the job_id, unless you were just wanting to get a count of all the jobs that were using resources of a particular type etc.).
2. If we are going to proceed as the tests added (and comments previously added) suggest, then should `attrs: []` return everything like `attrs: nil` does, instead of `job_id, status` like it does?
3. Should the required fields actually be [:job_id, :state_compact, :array_job_task_id]?

MorganRodgers:
2. and 3. I believe that we should always return at least the minimum viable set of attributes. Returning an empty array would be a big performance increase that would provide the user no value, so I do not see the point. I do not think that `array_job_task_id` is absolutely critical.

MorganRodgers:
#141 fixes #134.

PR #141, opened 2019-05-02
Fixes for Slurm adapter

MorganRodgers:
- Fix errors associated with `info_all_each` `attrs` not being `nil`
- Fix error associated with the record separator not being used to separate records
- Fix error associated with TRES bug

Hurray for being able to perform tests against a live system!


MorganRodgers:
@ericfranz I want to write a test for the difference between a 'plain' `squeue` and `squeue -M $cluster_name`, but I am having trouble. Do you have an example of `squeue -M` output?

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -150,7 +151,7 @@ def squeue_required_fields
           # -u <user_list> https://slurm.schedmd.com/squeue.html
           def squeue_args(id: "", options: [])
             args  = ["--all", "--states=all", "--noconvert"]
-            args += ["-o", "#{options.join(UNIT_SEPARATOR)}"]
+            args += ["-o", "#{RECORD_SEPARATOR}#{options.join(UNIT_SEPARATOR)}"]
```
MorganRodgers: Note that I had to move the record separator to the front to fix an error with `TRES`.

Issue #142, opened 2019-05-06
Don't Throw NoMethodError from OodCore::Job::ArrayIds


MorganRodgers:
Currently initializing `OodCore::Job::ArrayIds` with a `nil` or empty string throws when split is called on a `nil`. That should be handled.

PR #143, opened 2019-05-07
OodCore::Job::ArrayIds produces an empty array for invalid array requ…

MorganRodgers:
…ests

### DIFF
--- a/lib/ood_core/job/array_ids.rb
+++ b/lib/ood_core/job/array_ids.rb
@@ -1,19 +1,24 @@
 # Builds a sorted array of job ids given a job array spec string
-# 
+#
 # Job array spec strings:
 #   1         Single id
 #   1-10      Range
 #   1-10:2    Range with step
 #   1-10,13   Compound (range with single id)
-#   
+#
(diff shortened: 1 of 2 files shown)


MorganRodgers:
Fixes #142.

ericfranz:
Just a thought. Some of the complexity here is due to two things:

1. you can have invalid characters/states in each step, so multiple `raise Error`
2. you can have a range, an integer, or a range with a step, so multiple code paths for each

For the first, a more complex regex could possibly validate the entire string, and split and pull out relevant portions of each match, removing 

For the second, you could always assume there will be a range with a step. For example, if all you have is the integer 1, you could do `(1..1).step(1).to_a` to get `[1]`. So the default step would be 1, and then default upper range would be the same number as the lower.

Still doesn't capture the case where the range is invalid i.e. `2-1` instead of `1-2` or step is out of bounds `1-2:5`.

Those are refactoring suggestions that could wait. This PR has code that works and the tests to validate them.

MorganRodgers:
Note #144.

PR #146, opened 2019-05-10
Replace job array spec parser with Parslet::Parser

### LINKED ISSUE #144: Consider using regex to validate and parse job array spec
@ericfranz points out that a well crafted regex could potentially simplify the code for parsing and validating array specs in `OodCore::Job::ArrayIds`. I came up with the follow test:

```ruby
# Validates
# Contains only
# At least once
#   A number (start)
#     Optionally followed by a hyphen and a second number (stop)
#       Optionally followed by a percent and a third number (step)
#   Optionally followed by a comma
validate = /^(?:(?<start>[0-9]+)(?:-(?<stop>[0-9]+)(?:\%(?<step>[0-9]+))?)?,?)+$/
# Extracts
# A number (start)
#   Optionally followed by a hyphen and a second number (stop)
#     Optionally followed by a percent and a third number (step)
# Optionally followed by a comma
extractor = /(?:(?<start>[0-9]+)(?:-(?<stop>[0-9]+)(?:\%(?<step>[0-9]+))?)?,?)/

values = [
  # Valid
  '1',
  '1-4',
  '1-2,4',
  '1,2,3,4',
  '1-2,3-4,5',
  '1-4%2',
  '0-0%3',
  # Not valid
  '1,,2',
  'nil',
  '2 - 1 = 1',
  '1!@-#$%^&*()DHKI%^&*(@)#@_!-='
]

for value in values
  if validate =~ value
    p value.scan(extractor)
  else
    p "#{value} is NOT valid"
  end
end
```

Given that ArrayIds is used in the Grid Engine `qstat` handler we want to ensure that a regex based approach is at least as performant as the current implementation, in addition to being easier to understand.

MorganRodgers:
The current implementation of the job spec array parser is valid enough for production use although not perfect ([it fails this test](https://github.com/OSC/ood_core/blob/try-parslet/spec/job/array_ids_spec.rb#L88-L92)). @ericfranz suggested replacing the current implementation with something based on regular expressions, which is doable but not very readable.

An alternative to both the current implementation and a regex based approach is to use a [parsing expression grammar](https://en.wikipedia.org/wiki/Parsing_expression_grammar) framework like [Parselet](http://kschiess.github.io/parslet/). Using Parslet allows us to redefine the parsing and validating section of our code like so:

```ruby
class ArraySpecParser < Parslet::Parser
    rule(:integer) { match('[0-9]').repeat(1) }
    rule(:max_concurrent) { str('%') >> integer }
    rule(:range) { integer.as(:start) >> str('-') >> integer.as(:stop) }
    rule(:stepped_range) { range >> str(':') >> integer.as(:step) }
    rule(:component) { stepped_range | range | integer.as(:start) }
    rule(:additional_component) { str(',') >> component }
    rule(:array_spec) { component >> additional_component.repeat >> max_concurrent.maybe }
    root(:array_spec)
end

ArraySpecParser.new.parse('1-10:2%5')
# {:start => "1", :stop => "10", :step => "5"}
ArraySpecParser.new.parse('1-4,2')
# [{:start => "1", :stop => "10"}, {:start => "1"}]
ArraySpecParser.new.parse(nil)
# ArgumentError
ArraySpecParser.new.parse('this wont parse')
# Parslet::ParseFailed
```

There is a performance penalty to switching to Parslet with the Parslet implementation taking on average an additional 0.0027 seconds per run of the 13 out of 14 tests that master passes (each `bundle exec rspec spec/job/array_ids_spec.rb` was run 10 times). I am not convinced that this penalty is a deal breaker for us. I have not compared the Parslet implementation against one that is regex-based. Of course there is also the addition of a new dependency to consider. In general I think that we should prefer using something like Parslet over complex regular expressions when writing parsers unless we have a compelling reason to given the increase in clarity.

Fixes #144.

### DIFF
--- a/lib/ood_core/job/array_ids.rb
+++ b/lib/ood_core/job/array_ids.rb
@@ -1,3 +1,6 @@
+require "ood_core/refinements/array_extensions"
+require "parslet"
+
 # Builds a sorted array of job ids given a job array spec string
 #
 # Job array spec strings:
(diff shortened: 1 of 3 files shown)


ericfranz:
So my first thought was "YES!". But I have two follow up questions:

1. `1-4%2` is valid. Why is it necessary to flag `1-4%2,` as invalid?
2. This code uses regex and I think is simpler to understand than the above and passes all the tests. The code below can be modified to be more readable too. It has 3 clear captures of integers and uses those 3 captures right below via Ruby's `$1, $2, $3`.

```ruby
module OodCore
  module Job
    class ArrayIds
      attr_reader :spec_string

      def initialize(spec_string)
        @spec_string = spec_string
      end

      def ids
        @ids ||= parse_spec_string(spec_string)
      end

      protected

      def parse_spec_string(spec_string)
        return [] unless spec_string

        rx = /^(\d+)-?(\d+)?:?(\d+)?%?\d*$/
        spec_string.split(',').map do |spec|
          if rx =~ spec
            start = ($1 || 1).to_i
            finish = ($2 || start).to_i
            step = ($3 || 1).to_i
            (start..finish).step(step).to_a
          end
        end.compact.flatten
      end
    end
  end
end
```

I did not test the cost of doing this compared to using `Parslet::Parser`. This code could be made more readable or more performant. For example, using reduce in place of the map.compact.flatten would be faster.

ericfranz:
This passes the test as well and is likely the faster option:

```ruby
      def parse_spec_string(spec_string)
        return [] unless spec_string

        rx = /^(\d+)-?(\d+)?:?(\d+)?%?\d*$/
        spec_string.split(',').reduce([]) do |ids, spec|
          if rx =~ spec
            start = ($1 || 1).to_i
            finish = ($2 || start).to_i
            step = ($3 || 1).to_i
            ids.concat (start..finish).step(step).to_a
          end

          ids
        end
      end
```

ericfranz:
Alternative shared https://github.com/OSC/ood_core/pull/147

MorganRodgers:
Why would I call out `1-4%2,` as being invalid? Well, because it is: passing `1-4%2,` to Torque's `qsub` returns `Malformed array group` as an error. So by having a test for that particular input we give ourselves the option of not only parsing the array spec but also validating it. While none of the current implementations have an interface for reporting invalid array specs that might be a worthwhile feature.

In particular if we wanted to move from a generic array spec to adapter specific spec validations then I think Parslet would be a nice way to go because the composability of its Atoms lend themselves to producing readable variations on a theme. In terms of performance in Ruby I think that regex is more performant than the PEG for this task. So PEG then lives or dies on the merits of being more readable (or not) than regex. If you are not convinced that we have a readability win then we can put this to rest.

Tangentially: in the hypothetical Rust-y future there's a nice PEG library called [nom (from the gentle intro to Rust)](https://stevedonovan.github.io/rust-gentle-intro/nom-intro.html) that looks interesting and appears to be a win over regex.

PR #147, opened 2019-05-11
Refactor array ids parsing with regex

ericfranz:
Alternative to https://github.com/OSC/ood_core/pull/146


ericfranz:
With this change it would seem that all we need is a single function: spec string in, ids out. So do we still need a class for this function to live, or might it belong somewhere else?

MorganRodgers:
This significantly simplifies the code which is worth it.

PR #148, opened 2019-07-17
OpenStack adapter

ericfranz:
Associated Vagrant image for testing this adapter with: https://github.com/OSC/openstack

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737331617042) by [Unito](https://www.unito.io)

### DIFF
--- a/bin/get_servers
+++ b/bin/get_servers
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+echo "curl $BASE_URI using token $TOKEN"
--- a/bin/get_servers
+++ b/bin/get_servers
@@ -1,6 +1,10 @@
-#!/bin/bash
- 
-curl -s -H "X-Auth-Token: $TOKEN" \
-http://$BASE_URI/compute/v2.1/servers \
-
-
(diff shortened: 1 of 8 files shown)


bin/get_servers
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+ 
+curl -s -H "X-Auth-Token: $TOKEN" \
+http://$BASE_URI/compute/v2.1/servers \
```
ericfranz: Maybe try this

```
{ 
  statuscode=$(curl --silent --output /dev/stderr --write-out "%{http_code}" -H "X-Auth-Token: $TOKEN" http://$BASE_URI/compute/v2.1/servers);
} 2>&1

# now echo $statuscode gives 200 or 404 or 500
if [ "$statuscode" -ne 200 ]; then exit "$statuscode"; fi
```

ericfranz: Instead this:

```
output=$( { statuscode=$(curl --silent --show-error --output /dev/stderr --write-out "%{http_code}" https://gsadgdsagoogle.com/foo); } 2>&1; )

if [ "$statuscode" -ne 200 ];
then
  echo $output >&2
  exit 1
else
  echo $output
fi
```

johrstrom:
I'm closing this as it is not going to progress beyond this point at this time.

PR #151, opened 2019-09-09
LinuxHost adapter for running user scripts / containers

MorganRodgers:
Supports:
- 'Walltime' implemented using `timeout`
- An explicit SSH host whitelist
- Changing directory to a workdir
- Discovery of running jobs using `tmux list-panes`
- IO redirection
- Removal of running jobs (using a pipeline to send SIGKILL to the sinit process)
- Setting environment variables (which occurs on top of a user's regular environment)
- Emailing one or more users on job start (effectively instantly) and job end so long as job end is caused by `timeout`, `SIGTERM` or `SIGINT`.

Does NOT support:
- Holding / releasing jobs
- List fork jobs by other users


MorganRodgers:
For reference until tests /specs are written I have been testing using the following:

```ruby
require 'ood_core'
require 'ood_core/job/adapters/fork'
require 'ood_core/job/adapters/fork/forker'

config = {
    :adapter => 'fork',
    :tmux_bin => '/usr/bin/tmux',
    :ssh_hosts => [
        # Attempts SSH to host attempted to confirm key hash which breaks SSH batch mode
        # 'owens-login01.hpc.osc.edu',

        'owens-login02.hpc.osc.edu',
        'owens-login03.hpc.osc.edu',
    ],
    :submit_host => 'owens.osc.edu',
    :max_timeout => 120,  # in seconds
    :debug => false
}

# forker = OodCore::Job::Adapters::Fork::Forker.new(**config)
adapter = OodCore::Job::Factory::build(config)

content = <<-HEREDOC
# Don't serve your ~/.ssh directory...
cd /users/PZS0002/mrodgers/dev/quincy || exit 1
# starts up a HTTP server on port 8000
# Reachable at https://ondemand-test.osc.edu/rnode/$HOST/8000/
# Note that the trailing slash is important
python -m SimpleHTTPServer
exit 0
HEREDOC

env = {
    'KEY' => 'value!',
    'HOMEBOI' => 'Rasputin'
}
wall_time = 360

script = OodCore::Job::Script.new(
    accounting_id: nil,
    args: nil,
    content: content,
    email: nil,
    email_on_started: nil,
    email_on_terminated: nil,
    error_path: '/users/PZS0002/mrodgers/stderr_from_fork.log',
    input_path: nil,
    job_array_request: nil,
    job_environment: env,
    job_name: nil,
    output_path: '/users/PZS0002/mrodgers/stdout_from_fork.log',
    native: nil,
    priority: nil,
    queue_name: nil,
    rerunnable: nil,
    reservation_id: nil,
    shell_path: nil,
    start_time: nil,
    submit_as_hold: nil,
    wall_time: wall_time,
    workdir: '/users/PZS0002/mrodgers/dev/ood_core'
)

p 'Submitting'
session_id = adapter.submit(script)
p session_id
# p 'Info single job'
# p adapter.info(session_id)
# p 'Info all'
# p adapter.info_all

# sleep 1

# p 'Delete'
# adapter.delete(session_id)
# p 'Info single job (after delete)'
# p adapter.info(session_id)
# p 'Info all (after delete)'
# p adapter.info_all

```

treydock:
Could you hold/resume by `kill -STOP` and `kill -CONT`? This is kind of how resource managers like SLURM support suspend preemption.

MorganRodgers:
> Could you hold/resume by `kill -STOP` and `kill -CONT`

I was thinking about using `SIGSTOP` as an option. What concerns me, although I haven't tried it, is how to detect stopped `tmux` sessions, and how to get a real PID from them to be able to signal the process. I haven't been able to get a PID using `tmux list-sessions -F '...format string here...'`.

johrstrom:
You extract the hostname out of the output, why not `$$`.  Besides that, if the `session_name` was a concatination of hostname + pid we may have an easier time keeping track of sessions we've started. 

It seems like the caller is going to have to keep track because this is stateless? I get why we do that for other adapters because they take care of managing things, but I feel like this adapter may have to do extra work to clean up after itself? We take a lot of behaviours for granted from slurm or torque, etc. IDK if we want to be extra cautious with this or have a policy of 'sheller beware, clean up after yourself'. 

In either case, I'm OK with this as it stands and the TODOs you've provided in your comments but I feel like we need to come up with a way to really battle test this to make sure we're not forking tons of processes only to leave them long afterwards (like what happens when the `kill` fails, when something zombies, or any other sort of weird things).

ericfranz:
ood_core is stateless

>this adapter may have to do extra work to clean up after itself

this is where I'm still interested to see the use Singularity images to start the processes in a PID namespace (I think you have to use a separate flag with Singularity)

ericfranz:
When we originally took over building the interactive apps we were just focused on TurboVNC and when we started vncserver we included a flag to kill itself after 5 minutes of inactivity. Since this is a fork adapter were you can literally just start things up immediately, and we are talking about doing this primarily for interactive work (i.e. using RStudio etc.) we should add the concept of killing itself after a certain amount of time of inactivity passes. I imagine that to do that we would need some type of sidecar proxy. Maybe a requirement for the service mesh...

MorganRodgers:
> You extract the hostname out of the output, why not `$$`. Besides that, if the `session_name` was a concatenation of hostname + pid we may have an easier time keeping track of sessions we've started.

Capturing `$$` is a great idea. To be able to do it we need to get it from inside the tmux session and not the process that launches that session which is ephemeral. I think it should be doable. An idea that I haven't explored at all is rolling our own persistence mechanism for tracking host / PID combinations.

In any case if we have a PID and a hostname then `session_name` can be something more meaningful to the user like their job name.

> It seems like the caller is going to have to keep track because this is stateless? I get why we do that for other adapters because they take care of managing things, but I feel like this adapter may have to do extra work to clean up after itself? We take a lot of behaviours for granted from slurm or torque, etc. IDK if we want to be extra cautious with this or have a policy of 'sheller beware, clean up after yourself'.

The caller does need to keep track, but part of the goal is to have a reliable timeout mechanism.

> I feel like we need to come up with a way to really battle test this

I couldn't agree more.

treydock:
For timeout or kill after period, can start commands with `timeout` command. I frequently use this to kill processes after a given number of seconds.

MorganRodgers:
Updates: I have rewritten the adapter to use Singularity

> For timeout or kill after period, can start commands with `timeout` command. I frequently use this to kill processes after a given number of seconds.

I am now using timeout to implement `wall_clock`.

> > Could you hold/resume by `kill -STOP` and `kill -CONT`
> 
> I was thinking about using `SIGSTOP` as an option. What concerns me, although I haven't tried it, is how to detect stopped `tmux` sessions, and how to get a real PID from them to be able to signal the process. I haven't been able to get a PID using `tmux list-sessions -F '...format string here...'`.

By switching from `tmux list-sessions` to `tmux list-panes` I can now get a PID. I am not going to implement `Adapter#hold` and `Adapter#release` for the time being, but it looks like it may be doable if we really want it.

@treydock please take a look at https://github.com/OSC/ood_core/blob/f2b4b555884415d33b3f1b883a540cacb5a16f01/lib/ood_core/job/adapters/fork/launcher.rb#L107-L113

I have argued the case for this with @ericfranz  and @johrstrom , but wanted your take on it as well. The problem is I get SSH-in-batchmode failures when `owens-login01` is selected by the round robin due to host checking. It occurs to me that `known_host` related failures could be a frequent source of complaints from users, and so it makes sense to provide the ability to skip strict host checking. The way I see it the current solution:
    - Defaults to performing strict host checking
    - Only impacts use of this adapter, because it does not change user's `~/.ssh/config`
    - Improves experience of using the adapter, especially in round robin login situations (like at OSC)

treydock:
I think those SSH options are good. Are you still getting errors at OSC with those settings? We use those settings with clustershell and pdsh which uses SSH to connect to compute and login:

clustershell
```
ssh_options: -oStrictHostKeyChecking=no -oUserKnownHostsFile=/dev/null -oLogLevel=ERROR
```
pdsh:
```
export PDSH_SSH_ARGS_APPEND="-oStrictHostKeyChecking=no -oUserKnownHostsFile=/dev/null -oLogLevel=ERROR"
```

I don't recall why I did `LogLevel` but also maybe useful if you're seeing warnings at the user level.

MorganRodgers:
> Are you still getting errors at OSC with those settings?

No, when those settings are used I do not have problems, I just wanted to get a sense of if they would be perceived as a bad idea by sysadmins. Thanks for looking at it!

lib/ood_core/job/adapters/fork/launcher.rb
```diff
@@ -0,0 +1,241 @@
...
+      hsh.delete('SINGULARITY_CONTAINER')
+    }.map{
+      |key, value| "export #{key}=#{Shellwords.escape(value)}"
+    }.sort.join("\n")
+  end
+
+  def singularity_image(script)
+    image = script.job_environment['SINGULARITY_CONTAINER']
+    (image) ? image : default_singularity_image
+  end
```
MorganRodgers: The environment variable `SINGULARITY_CONTAINER` is set by Singularity at run time to point to the image, but is not usually used to define the image-to-be-run as it is here. The alternative to using the environment would be to use the `native` property on the `OodCore::Job::Script`. I chose to use the environment because we are already using the environment to set `SINGULARITY_BINDPATH`, and because in general we discourage the use of `native`.

Thoughts?

johrstrom: Why is `native` discouraged? I can't really think of a strong argument for or against ENV variables in genera other than maybe it's not a good idea to overload other peoples' env variables? We then have to worry about when they may unset, if we want to retry for example.

MorganRodgers: To venture a guess; native is discouraged because by doing so the library implementor is encouraged to look for similarities between different schedulers/job-launch paradigms, and so make better use of our existing interface.

Thinking of the risk of overloading the meaning of an existing environment variable is good advice. As of Singularity 3.x the container env var is only set within a running container, and I unset it before the container starts, but that's not to say that the Singularity team doesn't own that variable, and so may change what it does in future releases.

MorganRodgers: To pick this back up, I am now leaning towards using `native` for image selection and specifying bindmounts. The reason is because at present the batch plugin interface does not expose the ability to set environment variables, but it does expose `script.native`.

lib/ood_core/job/adapters/linux_host/launcher.rb
```diff
@@ -0,0 +1,262 @@
...
+    # discover tmux sessions
+    # discover the PID for the given session
+    # get all children of that PID
+    # find the sinit process for the tmux PID
+    # kill the sinit process
+    kill_cmd = <<~SCRIPT
+    kill $(pstree -p $(tmux list-panes -aF '#\{session_name} \#{pane_pid}' | grep '#{session_name}' | cut -f 2 -d ' ') | grep -Po 'sinit\\(\\d+' | grep -Po '[0-9]+')
+    SCRIPT
+
+    call(*cmd, stdin: kill_cmd)
```
MorganRodgers: As discussed in-person: this change from using `tmux kill-session` is necessary because user scripts with non-trivial activities (like desktops) were not dying when their parent tmux session died; killing the running container works as expected. Simpler ways of doing this are welcome.

Note also that this creates explicit system dependencies on:

- GNU `grep` (I'm going to rewrite one last time to remove the GNU / `-P` requirement)
- `pstree`
- `cut`

MorganRodgers: ... which brings the total system dependency list necessary to use this feature to:

- `cut`
- `grep`
- `pstree`
- `Singularity`
- `ssh`
- `tmux`

MorganRodgers: @treydock will you please review one more time? In particular the method to kill sessions changed significantly: https://github.com/OSC/ood_core/pull/151/files#diff-ef23b874dfc17f3318ec7a3985205bbcR69-R89

lib/ood_core/job/adapters/linux_host/launcher.rb
```diff
@@ -0,0 +1,262 @@
...
+  def stop_remote_session(session_name, hostname)
+    cmd = ssh_cmd(hostname)
+
+    # discover tmux sessions
+    # discover the PID for the given session
+    # get all children of that PID
+    # find the sinit process for the tmux PID
+    # kill the sinit process
+    kill_cmd = <<~SCRIPT
+    kill $(pstree -p $(tmux list-panes -aF '#\{session_name} \#{pane_pid}' | grep '#{session_name}' | cut -f 2 -d ' ') | grep -o 'sinit([[:digit:]]*' | grep -o '[[:digit:]]*')
```
treydock: There is a lot going on in one line. Just for readability maybe split each `$()` execution into separate variable:

```
PAIN_PIDS=$(tmux list-panes -aF '#\{session_name} \#{pane_pid}' | grep '#{session_name}' | cut -f 2 -d ' ')
PIDS=$(pstree -p "${PAIN_PIDS}" | grep -o 'sinit([[:digit:]]*' | grep -o '[[:digit:]]*')
kill "${PIDS}"
```
Might have to adjust double quotes around variables depending on what output looks like.

What do session names look like? Would it be necessary for `grep '#{session_name}'` to be `grep -e '^#{session_name}' just to ensure correct match?

Also is the output of each `$()` going to contain new lines? Will the commands they are passed to handle newlines? Like will you end up saying this?

```
kill 123
124
125
```

MorganRodgers: Session names are purely programmatic, and look like `launched_by_ondemand-$UUID`. The tmux pane format strings look a lot like Ruby string interpolation which is a little confusing hence the escapes in one `session_name` but not the second. Since the session name is interpolated into the heredoc by Ruby there should only ever be one match; so `kill` would only get a single PID.

treydock: What about the output of `pstree`? WIll that contain newlines or just one PID? I would assume more than one PID is output by `pstree` or does the greps after ensure just one PID is returned?

MorganRodgers: By looking for `sinit(` I'm able to cut it down to a single PID.

lib/ood_core/job/adapters/linux_host/launcher.rb
treydock: Looks fine to me.

MorganRodgers: Thanks.

Issue #150, opened 2019-09-06
Catch exceptions when reading bad cluster yml files


johrstrom:
We've gotten issues from users of this library (ourselves in the Ondemand dashboard) where exceptions are thrown from reading invalid yml files for cluster configs.

The offending read line is [here](https://github.com/OSC/ood_core/blob/869cd93878c30ae4fc9f469f062b9452d94b61b4/lib/ood_core/clusters.rb#L31) and the solution seems simply to catch (rescue in ruby parlance) the exception, spitting out the relevant message in logs. 
```ruby
          Pathname.glob(config.join("*.yml")).each do |p|
            CONFIG_VERSION.any? do |version|
              if cluster = YAML.safe_load(p.read).fetch(version, nil)
                clusters << Cluster.new(send("parse_#{version}", id: p.basename(".yml").to_s, cluster: cluster))
                true
              else
```

ericfranz:
This library currently has no logger available to it. If #86 were available, that could be used. But the problem with logging and returning an empty array is that it makes it more difficult for the code that uses the library to determine that an error occurred. All the user of this code would know is that there are no clusters available.

I remember when Jeremy first avoided fixing this. His point was that we shouldn't let the app start with a malformed cluster config - we should let it crash so the user immediately knows the problem. Now with the current version of OnDemand, I'm not sure if this error occurs early enough in the launching of the Dashboard app to cause Passenger to display that debug page, and users may still need to fish through logs to determine what happened.

Since this is a library I think there are two approaches preferable to logging an error and returning a Clusters object with a zero length of clusters.

1. rescue specific exceptions, like the `Psych::SyntaxError` or other errors, and re-raise an ood_core specific error
   - similar to [ConfigurationNotFound](https://github.com/OSC/ood_core/blob/869cd93878c30ae4fc9f469f062b9452d94b61b4/lib/ood_core/clusters.rb#L40) this could be a `ConfigurationMalformedError` error or `ConfigurationParsingError`
[...]

PR #152, opened 2019-09-20
Add cluster ACLs based on the cluster config's file permissions

MorganRodgers:
Fixes crash when a YAML file in the clusters.d is not readable by the user

### DIFF
--- a/lib/ood_core/clusters.rb
+++ b/lib/ood_core/clusters.rb
@@ -18,10 +18,12 @@ module OodCore
         config = Pathname.new(path.to_s).expand_path
 
         clusters = []
-        if config.file? && config.readable?
-          CONFIG_VERSION.any? do |version|
-            YAML.safe_load(config.read).fetch(version, {}).each do |k, v|
-              clusters << Cluster.new(send("parse_#{version}", id: k, cluster: v))
+        if config.file?
+          if config.readable?
+            CONFIG_VERSION.any? do |version|
+              YAML.safe_load(config.read).fetch(version, {}).each do |k, v|
+                clusters << Cluster.new(send("parse_#{version}", id: k, cluster: v))
+              end
             end
           end
         elsif config.directory?
@@ -32,7 +34,7 @@ module OodCore
               end
             end
           end
-        elsif ! config.exist?
+        else
           raise ConfigurationNotFound, "configuration file '#{config}' does not exist"
         end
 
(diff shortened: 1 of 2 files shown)


johrstrom:
I know Erik has thoughts on whether we should rescue errors in this library or bubble them up to higher layers.  I could see it both ways, but I think if we bubble it up it's easier to show the user what happened. And I feel like that's what we need to fix, communicating the failure to the user, and the potential for failure to the developer. Same with #150 actually. 

That said, if we do rescue errors here, we need to log or indicate somehow that they occurred.  Otherwise we leave people scratching their heads because nothing happened and also nothing was said to have failed when clearly something, somewhere did.

MorganRodgers:
In general I take the point about libraries bubbling errors up. I this case, limiting access to clusters using file system permissions is a feature admins might reasonably expect to just work, and so supporting that behavior outside of the library means that all the apps would need to handle the error which seems unnecessary. This issue has come up for me before where I thought that we controlled access to clusters using file system permissions or ACLs. The surprising behavior was when not only did that not work, but it crashed the Dashboard before Passenger had spun up enough to give me a nice error page:

```plaintext
500 Internal Server Error
If you are the administrator of this website, then please read this web application's log file and/or the web server's log file to find out what went wrong.
```

We could log `Errno::EACCES` if that is desired.

johrstrom:
Oh i see, so it's not really failing, it's actually expected behavior. OK cool.  As to logging, I think that's a precedent we're probably not able to set quite yet especially in this case.

ericfranz:
If a cluster config is not readable by the user it is okay to ignore the file without raising an exception or logging an error, since this is a feature - a normal case. We do not log errors to the Dashboard's log file every time a user accesses the dashboard just because an app happens to have restricted permissions on it. In other words, it wouldn't be an exceptional case - though I imagine it has happened before. There are other strategies we can take to help admins verify they intended to restrict the permissions on a cluster config file.

But using exception handling to avoid the crash is the wrong approach. Instead, check if the file is readable by the user. For example:

```ruby
  if config.file? && config.readable?

# ...

  elsif config.directory?
    Pathname.glob(config.join("*.yml")).select(&:file?).select(&:readable?).each do |p|
```

FWIW The original method is difficult to read and should be refactored. For example, there is no reason to ensure that the block passed to `any?` properly returns false if the clusters array is empty because we do nothing with that return value. Instead of building an array of clusters with each, we could probably do a map. We do not need the v1 config anymore either, though removal of v1 is technically a breaking change in ood_core.

PR #154, opened 2019-10-02
Adds a configuration class for use by all Ruby-based apps

MorganRodgers:
Added to ood_core because ood_appkit requires Rails which is not the base framework for all our apps.

* Supports programmatic configuration using ERB defaults to `/etc/ood/config/bin/erb_wrapper`
* Reacts well to options missing from the config

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201737436582662) by [Unito](https://www.unito.io)

### DIFF
--- a/lib/ood_core/app_config.rb
+++ b/lib/ood_core/app_config.rb
@@ -8,11 +8,12 @@ module OodCore
 
         attr_reader :raw_config
 
-        def initialize
+        def self.load_config(app_config_path=nil)
+            AppConfig.new(load_yaml(app_config_path) || {})
         end
 
-        def load_config
-            @raw_config = YAML.load(load_yaml) || {}
+        def initialize(opts={})
+            @raw_config = opts
         end
 
         def branding_bg_color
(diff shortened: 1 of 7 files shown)


lib/ood_core/app_config.rb
```diff
@@ -0,0 +1,168 @@
+require "ood_core/refinements/array_extensions"
+require 'pathname'
+require 'yaml'
+
+module OodCore
+    class AppConfig
```
ericfranz: Should this be a class or a module?

ericfranz: I guess as a class if you could do this:

```ruby
def initialize(raw_config)
  @raw_config = config
end
```

lib/ood_core/app_config.rb
```diff
@@ -0,0 +1,168 @@
...
+        using Refinements::ArrayExtensions
+
+        attr_reader :raw_config
+
+        def initialize
+        end
+
+        def load_config
+            @raw_config = YAML.load(load_yaml) || {}
+        end
```
ericfranz: `load_config` and `load_yaml` etc. should probably all be moved/grouped together as factory methods (as class methods or in a separate AppConfigFactory) and `initialize(raw_config)` so that to create the base config file you just do this: AppConfig.new({})

That could also help simplify some tests.

lib/ood_core/app_config.rb
```diff
@@ -0,0 +1,168 @@
+require "ood_core/refinements/array_extensions"
+require 'pathname'
+require 'yaml'
+
+module OodCore
+    class AppConfig
+        using Refinements::ArrayExtensions
+
+        attr_reader :raw_config
```
ericfranz: Why call it `raw_config` instead of just `config`?

lib/ood_core/app_config.rb
ericfranz: It isn't clear to me yet how app specific ConfigurationSingleton classes will work with this "base class" (subclass? decorator? possibly decorator) but if individual apps are provided the option to instantiate the class (by calling, for example, AppConfig.load_config to get the raw config hash) then apps could load the base config hash, load the config the previous way, do a merge, and instantiate the resulting Configuration singleton, providing:

1. backwards compatibility with older configuration options
2. preservation of app specific configuration that does not belong here (like dataroot, database_path, etc.)

lib/ood_core/app_config.rb
MorganRodgers: `raw_config` signifies that it has not been processed. No defaults are set and no string-paths have been converted to `Pathnames`.

lib/ood_core/app_config.rb
MorganRodgers: I think that merging Hashes is the way to go; how else would we want to do it when we get to supporting user defined configuration?

lib/ood_core/app_config.rb
```diff
@@ -0,0 +1,168 @@
...
+            raw_config['whitelist_path'] || []
+        end
+
+        class Error < StandardError; end
+
+        private
+
+        def load_yaml
+            erb_wrapper_path = Pathname.new(ENV['OOD_CONFIG_ERB_WRAPPER'] || '/etc/ood/config/bin/erb_wrapper')
+            app_config_path = Pathname.new(ENV['OOD_APP_CONFIG'] || '/etc/ood/config/app_config.yml')
```
MorganRodgers: > If we are supporting ERB as suggested in https://github.com/OSC/ood_core/pull/154, should we add the ERB here to build the list of favorites?

_Originally posted by @ericfranz in https://github.com/OSC/osc-ood-config/pull/39

Note that it is really only suggested to use ERB. So long as it is executable (and works in the calling environment) and outputs YAML to stdout there is no reason it couldn't be written in another language.

This section needs more work. There needs to be a fallback if neither `OOD_CONFIG_ERB_WRAPPER` is set or `/etc/ood/config/bin/erb_wrapper` does not exist.

lib/ood_core/app_config.rb
MorganRodgers: Part of the idea was to be able to get at the data before it was transformed in case the transformation results in lost information.

ericfranz: Sounds good

johrstrom:
I'm going close this as it's not going to make progress anytime soon.

Issue #153, opened 2019-09-30
Assuming netcat exists can cause problems


ericfranz:
See: https://github.com/OSC/ood_core/blob/a87f59ea9d24cdcbe505c9ab859bcb120bae9951/lib/ood_core/batch_connect/template.rb#L120-L126

The assumption here is that netcat is available. If it is not (because the command doesn't exist or the user is not allowed to execute the command) the result will be nonzero but will be reported as the port not yet being in use.

If we are going to use netcat, we should check for its existence and handle the error case that, then use fallback solutions.

Or not use netcat at all. See https://listsprd.osu.edu/pipermail/ood-users/2018-October/000269.html for discussion and some options and Trey's comment:

>I’ve used this trick in the past to check if ports are available when bootstrapping MySQL instances in jobs that shared the same node:
> 
>    if (: < /dev/tcp/127.0.0.1/${MYSQL_PORT}) 2>/dev/null; then
>        echo "MySQL port ${MYSQL_PORT} already taken."
>        return 1
>    fi
>

ericfranz:
On solution if `lsof` is available on the compute nodes to add to the before.sh.erb before `find_port` is executed:

```bash
# override
port_used () {
  local port="${1#*:}"
  local host=$((expr "${1}" : '\\(.*\\):' || echo "localhost") | awk 'END{print $NF}')
  lsof -i :"${port}" &> /dev/null
}
export -f port_used
```

ericfranz:
If there is a reliable way to check if the pseudo-device /dev/tcp is available, we would check if that exists first, then use it if it does; if not then try something else.

ericfranz:
Addressed by #159

PR #159, opened 2019-11-25
fail fast if nc is not installed and indicate that it isn't installed

johrstrom:
Clearly most sites don't already have nc installed.  This will actually tell them that it isn't installed.

I was able to test with, commenting out nc and adding the 'notfound' command.

```bash
    #nc -w 2 "${host}" "${port}" < /dev/null > /dev/null 2>&1
    notfound >/dev/null 2>&1
```

### DIFF
--- a/lib/ood_core/batch_connect/template.rb
+++ b/lib/ood_core/batch_connect/template.rb
@@ -116,27 +116,49 @@ module OodCore
                   shuf -i ${1}-${2} -n 1
                 }
                 export -f random_number
+                export python_socket_cmd="import socket; socket.socket().connect(('$1',$2))"
+
+                port_used_python() {
+                  python -c "$python_socket_cmd" >/dev/null 2>&1
+                }
+
+                port_used_python3() {
+                  python3 -c "$python_socket_cmd" >/dev/null 2>&1
+                }
+
+                port_used_nc(){
+                  nc -w 2 "$1" "$2" < /dev/null > /dev/null 2>&1
+                }
+
+                port_used_lsof(){
+                  lsof -i :"$2" >/dev/null 2>&1
+                }
+
+                port_used_bash(){
+                  local bash_supported=$(strings /bin/bash 2>/dev/null | grep tcp)
+                  if [ "$bash_supported" == "/dev/tcp/*/*" ]; then
+                    (: < /dev/tcp/$1/$2) >/dev/null 2>&1
+                  else
+                    return 127
+                  fi
+                }
 
                 # Check if port $1 is in use
                 port_used () {
                   local port="${1#*:}"
                   local host=$((expr "${1}" : '\\(.*\\):' || echo "localhost") | awk 'END{print $NF}')
-                  local python_cmd="import socket; socket.socket().connect(('$host',$port))"
-                  local bash_supported=$(strings /bin/bash | grep tcp)
+                  local port_strategies=(port_used_nc port_used_lsof port_used_bash port_used_python port_used_python3)
 
-                  if nc -h >/dev/null 2>&1; then
-                    nc -w 2 "${host}" "${port}" < /dev/null > /dev/null 2>&1
-                  elif lsof -h >/dev/null 2>&1; then
-                    lsof -i :"${port}" >/dev/null 2>&1
-                  elif python -h >/dev/null 2>&1; then
-                    python -c "$python_cmd" >/dev/null 2>&1
-                  elif python3 -h >/dev/null 2>&1; then
-                    python3 -c "$python_cmd" >/dev/null 2>&1
-                  elif [ "$bash_supported" == "/dev/tcp/*/*" ]; then
-                    (: < /dev/tcp/${host}/${port}) >/dev/null 2>&1
-                  else
-                    return 127
-                  fi
+                  for strategy in ${port_strategies[@]};
+                  do
+                    $strategy $host $port
+                    status=$?
+                    if [[ "$status" == "0" ]] || [[ "$status" == "1" ]]; then
+                      return $status
+                    fi
+                  done
+
+                  return 127
                 }
                 export -f port_used
 
--- a/lib/ood_core/batch_connect/template.rb
+++ b/lib/ood_core/batch_connect/template.rb
@@ -121,7 +121,13 @@ module OodCore
                 port_used () {
                   local port="${1#*:}"
                   local host=$((expr "${1}" : '\\(.*\\):' || echo "localhost") | awk 'END{print $NF}')
-                  nc -w 2 "${host}" "${port}" < /dev/null > /dev/null 2>&1
+                  if command -v nc >/dev/null 2>&1; then
+                    nc -w 2 "${host}" "${port}" < /dev/null > /dev/null 2>&1
+                  elif command -v lsof >/dev/null 2>&1; then
+                    lsof -i :"${port}" >/dev/null 2>&1
+                  else
+                    return 127
+                  fi
                 }
                 export -f port_used
 
@@ -148,7 +154,8 @@ module OodCore
                     if [ "$port_status" == "0" ]; then
                       return 0
                     elif [ "$port_status" == "127" ]; then
-                       echo "nc command not found, please install it! exiting 127"
+                       echo "command to find port not found, please install it! exiting 127"
+                       echo "command options are lsof or nc"
                        return 127
                     fi
                     sleep 0.5
--- a/lib/ood_core/batch_connect/template.rb
+++ b/lib/ood_core/batch_connect/template.rb
@@ -121,14 +121,19 @@ module OodCore
                 port_used () {
                   local port="${1#*:}"
                   local host=$((expr "${1}" : '\\(.*\\):' || echo "localhost") | awk 'END{print $NF}')
-                  if command -v nc >/dev/null 2>&1; then
+                  local python_cmd="import socket; socket.socket().connect(('$host',$port))"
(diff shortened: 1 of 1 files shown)


ericfranz:
Instead I’d rather see nc be only one strategy. If bash supports the /dev/tcp device to use it. If lsof is available use it. If python is available use it. Etc.

ericfranz:
See https://github.com/OSC/ood_core/issues/153

johrstrom:
Ok this now has `nc`, `lsof`, `bash /net/tcp`, and `python` checks in that order. If there's still nothing defined it'll say so and say that it checked those 4 options.

lib/ood_core/batch_connect/template.rb
```diff
@@ -121,7 +121,17 @@ def bash_helpers
                 port_used () {
                   local port="${1#*:}"
                   local host=$((expr "${1}" : '\\(.*\\):' || echo "localhost") | awk 'END{print $NF}')
-                  nc -w 2 "${host}" "${port}" < /dev/null &> /dev/null
+                  if command -v nc >/dev/null 2>&1; then
+                    nc -w 2 "${host}" "${port}" < /dev/null > /dev/null 2>&1
+                  elif command -v lsof >/dev/null 2>&1; then
+                    lsof -i :"${port}" >/dev/null 2>&1
+                  elif [ -n "$SHELL" ] && [ "$SHELL" = "/bin/bash" ]; then
```
MorganRodgers: Eric expressed concern about `/dev/tcp` not being available for every version of Bash. Is that concern valid? If so this does not test to ensure that the version of Bash is compatible with `/dev/tcp`.

johrstrom: Not sure, I just checked on the 3.0 container and it existed there and 3.0 is from 2004.  We could make stronger checks.  It would throw `No such file or directory` if it didn't exist.

ericfranz: /dev/tcp needs to be checked for sure. AFAIK centos6 bash doesn’t have this even though we dropped support for centos6 on the webnode.

Also these check for the existence of the command and the assume any non zero status of the command means the port is not in use. This is also a problem. We should check for the specific exit status that indicates the port in use or not. If another exit status occurs, we should assume that means the method being used will not work and move onto the next method to check.

An example of the importance of this is VT where nc exists as a command but when users execute it the stderr gets the output something like “You do not have permission to run this command”.

MorganRodgers: For testing Bash how about this:

```bash
function can_use_bash() {
    ( exec 3<>/dev/tcp/www.google.com/80 ) 2>/dev/null
    "$?"
}
```

johrstrom: Yea i keep forgetting this is open.  

I'm not sure if we can get the level of error handling through return codes.  I just tried a bad device `/dev/tcp1` and it returned 1, then I tried an unknown/bad host `/dev/tcp/www.jeffohrstrom.biz/80` and got 1 again. 

I did get different error messages, and that's probably what we'll have to go on, actually grepping for `No such file or directory` and `Connection refused|Invalid argument` (`/dev/tcp/localhost/80` gave me connection refused).

MorganRodgers: That might make more sense. I don't love pointing at Google just to test if Bash is configured to use TCP. I'd love some way to just ask Bash if it supports a given feature. There's (`readelf`)[https://linux.die.net/man/1/readelf], which we could use like this: `readelf -a /usr/bin/bash | grep have_devfd`, but I am not sure that `have_devfd` is the correct symbol to look for. It seems promising...

MorganRodgers: I think we may have a winner: https://github.com/bminor/bash/blob/64447609994bfddeef1061948022c074093e9a9f/lib/sh/netopen.c#L1-L3

So the `readelf` call should really be: `readelf -a /usr/bin/bash | grep netopen`

MorganRodgers: Judging by the binary produced by the source available from GNU (http://ftp.gnu.org/gnu/bash/) it appears that `--enable-net-redirections` is the default for Bash. In Bash 5.x the configuration option ` --disable-net-redirections` is needed to produce Bash without `/dev/tcp/*/*` capability.

I have confirmed that the feature can be detected in a Bash binary using: `strings /bin/bash | grep tcp`. `strings` is present in base CentOS containers when even `which` is not, and so seems like a safe dependency.

ericfranz:
What is the rationale of a904c03dd76d6dab1bff3d9cd34da55981a1b5be?

johrstrom:
You were right about `command` it'll still return 0 even when the command isn't executable. If the command isn't executable, like you don't have the right permissions/aren't in the right group, it'll return 126. If it doesn't exist it'll return 127. so `foo -h` will catch both problems, it'll only return 0 when it exists and you can run it.

As to changing the order for python, I just python is a fairly safe choice.

ericfranz:
The current logic still makes me uncomfortable. If `command -h` returns with a zero exit status then port_used will use the given strategy to determine if the port is in use, regardless of the exit status of that strategy.

Unfortunately I wasn't able to verify that the exit status from nc in VT's case returned a value not 1 or 0 as that is my specific use case for another situation besides the command not found that could be problematic.

It is too bad all these tools use 1 exit status to indicate port is not in use, and 1 for other error cases, so we can cleanly differentiate between the two. But in the rare case one of these commands returned another error status, and we could handle that and move to another strategy instead, that might be preferable.

Is this better?

```
port_used_via_bash_strategy () {
  (: < /dev/tcp/$1}/$2) >/dev/null 2>&1
}

port_used_via_nc_strategy () {
  nc -w 2 "$1" "$2" < /dev/null > /dev/null 2>&1
}

port_used () {
  local port="${1#*:}"
  local host=$((expr "${1}" : '\\(.*\\):' || echo "localhost") | awk 'END{print $NF}')

  port_used_via_nc_strategy
  port_status=$?
  if [ "$port_status" == "0" ] || [ "$port_status" == "1" ]; then
    return $port_status
  fi

  port_used_via_bash_strategy
  port_status=$?
  if [ "$port_status" == "0" ] || [ "$port_status" == "1" ]; then
    return $port_status
  fi

  # etc

  return 127;
}
```

This normally makes me shudder (all the return statements etc.) and there may be a better way to structure the code, but the idea here is we are explicitly testing for the return codes that are intended to indicate whether or not the port is in use.

Of course, if all of these commands use 1 exit status to indicate port not in use, or all other error codes, then there is no hope and my proposal won't have any effect. This is the unfortunate case for the python code, where the exit status is 1 regardless of whether the port connection fails or the python code is malformed. Unfortunately, doing anything more (like catching an exception and exiting with another custom status) would result in a string of code that is not compatible between Python 2 and Python 3. And its hard to imagine a site that doesn't have nc AND lsof though I imagine they exist. Maybe the bash strategy should be preferred over the python ones? I don't know.

ericfranz:
>You were right about command it'll still return 0 even when the command isn't executable. If the command isn't executable, like you don't have the right permissions/aren't in the right group, it'll return 126. If it doesn't exist it'll return 127. so foo -h will catch both problems, it'll only return 0 when it exists and you can run it.

This PR has 9 commits. When it is all said and done, we will probably only have 1 "change" so sqashing these commits into 1 commit would be appropriate. At that time, the body of the commit message should explain the rationale for using -h instead of command -v just as you have explained it here, as this is very valuable.

johrstrom:
What about this?  Slightly cleaner, though it checks all strategies against 0 or 1 and doesn't really leave room for a specific strategy throwing a specific error. 

```bash
port_used_lsof(){
  lsof -i :"$2" >/dev/null 2>&1
}

# Check if port $1 is in use
port_used () {
  local port="${1#*:}"
  local host=$((expr "${1}" : '\\(.*\\):' || echo "localhost") | awk 'END{print $NF}')
  local port_strategies=(port_used_nc port_used_lsof port_used_bash port_used_python port_used_python3)

  for strategy in ${port_strategies[@]};
  do
    $strategy $host $port
    status=$?
    if [[ "$status" == "0" ]] || [[ "$status" == "1" ]]; then
      return $status
    fi
  done

  return 127
}
```

ericfranz:
Works for me

lib/ood_core/batch_connect/template.rb
```diff
@@ -116,27 +116,49 @@ def bash_helpers
                   shuf -i ${1}-${2} -n 1
                 }
                 export -f random_number
+                export python_socket_cmd="import socket; socket.socket().connect(('$1',$2))"
```
ericfranz: I don't think this actually works as intended. Instead of exporting, just duplicate the code between the two python methods.

```sh
port_used_python() {
  python -c "import socket; socket.socket().connect(('$1',$2))"
}
port_used_python3() {
  python3 -c "import socket; socket.socket().connect(('$1',$2))"
}
```

You can see how this doesn't work:

```
efranz@owens-login01:~$ export python_socket_cmd="print('$1', $2)"
efranz@owens-login01:~$ port_used_python() { python -c "$python_socket_cmd"; }
efranz@owens-login01:~$ port_used_python localhost 8080
('',)
efranz@owens-login01:~$ port_used_python() { python -c "print('$1', $2)"; }
efranz@owens-login01:~$ port_used_python localhost 8080
('localhost', 8080)
```

johrstrom:
I'm fine to do that, but I tested all of these as I went along in a shell and it worked. You can see for yourself at `/tmp/test_python.sh` on owens login1.

PR #162, opened 2019-12-06
Add directive_prefix to all adapters

### LINKED ISSUE #161: Return resource manager directive prefixes
A [new job validator](https://github.com/OSC/ondemand/pull/310) is added to the job composer to filter eligible scripts to set as the job script. One of the filtering criteria is if the first 1000 bytes of the file contains `#PBS` or `#SBATCH` or `#BSUB` or `#$`. This should ultimately be handled by ood_core by adding a method to return the "script_directive_prefix" which will return these strings to job adapters.

MorganRodgers:
Adapters which do not support directive_prefix (such as the LinuxHost) are expected to raise JobAdapterError instead of NotImplementedError.

Fixes #161.

### DIFF
--- a/lib/ood_core/job/adapter.rb
+++ b/lib/ood_core/job/adapter.rb
@@ -36,7 +36,7 @@ module OodCore
       # Retrieve info for all jobs from the resource manager
       # @abstract Subclass is expected to implement {#info_all}
       # @raise [NotImplementedError] if subclass did not define {#info_all}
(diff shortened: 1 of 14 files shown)


ericfranz:
Actually for those that do not have one I’d rather it return nil or “”

MorganRodgers:
`nil` would be more semantic than empty string.

PR #163, opened 2019-12-10
Copy environment

### LINKED ISSUE #158: Slurm adapter: srun cannot see $PATH/other env vars when submitted through job composer
Let me start off by stating I'm honestly not sure if this is an OOD issue or a Slurm issue. I'm starting but submitting the issue with OOD because everything still works fine outside of Job Composer. We simultaneously upgraded both Slurm (18 to 19) and OOD (to 1.6.20/1.35.3), so it's tough to say for certain, but everything was working fine before the upgrades.

In a nutshell: using job-composer (and only using job-composer) to submit a batch script will change the behavior of `srun ...` lines within the script: programs executed from srun have a very different set of environment variables (and no $PATH whatsoever) than programs/commands within the same script that are not prefixed with srun.  A workaround is to always explicitly use `srun --export=all ...`  but this is not necessary when `sbatch`ing from any command line (including OOD's builtin terminal) and also seems contrary to slurm's documentation for srun:

> –export=<environment variables [ALL] | NONE>
Identify which environment variables are propagated to the launched application. By default, all are propagated.

This is probably best demonstrated by example, so take a peek at the below output taken from running the attached batch file through Job Composer:

```
OnDemand version: v1.6.20 | Dashboard version: v1.35.3
Slurm version: slurm 19.05.3-2


EXPECTATION: srun lines have same access to environment
             vars (esp PATH!) as lines not using srun.

** module load perl    # or any binary in PATH ...

** /usr/bin/which perl ...
/packages/perl/5.28.1/bin/perl

** srun --export=ALL /usr/bin/which perl ...
/packages/perl/5.28.1/bin/perl

** srun /usr/bin/which perl ...
/usr/bin/which: no perl in ((null))
srun: error: cn53: task 0: Exited with exit code 1
#         ^---- REALITY: non-srun command works as expected...
#                        BUT the srun command fails.

EXPECTATION: srun defaults to --export=ALL behavior as
             is stated in the srun docs/man-page.

** /bin/env | grep -vc SLURM     # count non-slurm env vars ...
** /bin/env | grep ^PATH | grep -o ':' | wc -l    # count PATH elements ...
35
10

** srun --export=ALL /bin/env | grep -vc SLURM ...
** srun --export=ALL /bin/env | grep ^PATH | grep -o ':' | wc -l ...
40
10

** srun /bin/env | grep -vc SLURM ...
** srun /bin/env | grep ^PATH | grep -o ':' | wc -l ...
6
0
# ^---- REALITY: non-srun command works as expected...
#                BUT the srun command does not reflect the greater sbatch job-env.
```

This all seems pretty related to issue #109 and this thread https://listsprd.osu.edu/pipermail/ood-users/2018-December/thread.html#329 ... but not closely enough for me to immediately make use of either for a solution.

sbatch source:
[ood-srun-example.txt](https://github.com/OSC/ood_core/files/3810640/ood-srun-example.txt)

MorganRodgers:
Fixes #158.

### DIFF
--- a/lib/ood_core/job/adapters/lsf/helper.rb
+++ b/lib/ood_core/job/adapters/lsf/helper.rb
@@ -92,6 +92,14 @@ class OodCore::Job::Adapters::Lsf::Helper
     args += ["-W", (script.wall_time / 60).to_i] unless script.wall_time.nil?
     args += ["-L", script.shell_path.to_s] unless script.shell_path.nil?
 
+    # Skip if nil
+    env = script.job_environment || {}
+    if script.copy_environment?
+      args += ["-env", (["all"] + env.keys).join(",")]
+    elsif script.copy_environment? == false
+      args += ["-env", (["none"] + env.keys).join(",")]
+    end
+
     # input and output files
     args += ["-i", script.input_path] unless script.input_path.nil?
     args += ["-o", script.output_path] unless script.output_path.nil?
--- a/lib/ood_core/job/adapters/lsf/helper.rb
+++ b/lib/ood_core/job/adapters/lsf/helper.rb
@@ -92,8 +92,9 @@ class OodCore::Job::Adapters::Lsf::Helper
     args += ["-W", (script.wall_time / 60).to_i] unless script.wall_time.nil?
     args += ["-L", script.shell_path.to_s] unless script.shell_path.nil?
 
-    # Skip if nil
+    # environment
     env = script.job_environment || {}
+    # To preserve pre-existing behavior we only act when true or false, when nil we do nothing
     if script.copy_environment?
       args += ["-env", (["all"] + env.keys).join(",")]
     elsif script.copy_environment? == false
@@ -111,9 +112,6 @@ class OodCore::Job::Adapters::Lsf::Helper
     args += ["-u", script.email.join(",")] unless script.email.nil? || script.email.empty?
 
     args += script.native unless script.native.nil?
-
-    # environment
-    env = script.job_environment || {}
 
     {args: args, env: env}
   end
--- a/lib/ood_core/job/adapters/pbspro.rb
+++ b/lib/ood_core/job/adapters/pbspro.rb
@@ -261,6 +261,7 @@ module OodCore
           # Set environment variables
           envvars = script.job_environment.to_h
           args += ["-v", envvars.map{|k,v| "#{k}=#{v}"}.join(",")] unless envvars.empty?
+          args += ["-V"] if script.copy_environment?
 
           # If error_path is not specified we join stdout & stderr (as this
           # mimics what the other resource managers do)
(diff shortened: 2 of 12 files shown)


johrstrom:
Do we also need to populate the environment? That is, we call some open3 api with an environment hash, in lsf it's [`env = script.job_environment || {}`](https://github.com/OSC/ood_core/blob/master/lib/ood_core/job/adapters/lsf/helper.rb#L108) that ends up being the actual env being passed into [open3.capture3](https://github.com/OSC/ood_core/blob/master/lib/ood_core/job/adapters/lsf/batch.rb#L146).  

So my question is, if someone sets this flag is it sufficient to just populate the command, or do we actually need to env variables to this hash so they get passed into open3.capture3's env?

MorganRodgers:
I believe in every case the environment is being passed to the `call` implementation which uses `open3.capture3`.

lib/ood_core/job/adapters/lsf/helper.rb
```diff
@@ -92,6 +92,14 @@ def batch_submit_args(script, after: [], afterok: [], afternotok: [], afterany:
     args += ["-W", (script.wall_time / 60).to_i] unless script.wall_time.nil?
     args += ["-L", script.shell_path.to_s] unless script.shell_path.nil?
 
+    # Skip if nil
+    env = script.job_environment || {}
```
ericfranz: This line  (`env = script.job_environment || {}`)is duplicate of line 116 below. I guess delete the one below?

johrstrom:
I think I found the answer from the open3 and process docs, I think that env is additive to the current ENV, not an out right replacement, so I think I answered my own question in that we only need to modify `args` unless there's somehting specific to `env` we need to deal with like in the case of slurm.

lib/ood_core/job/adapters/lsf/helper.rb
```diff
@@ -92,6 +92,14 @@ def batch_submit_args(script, after: [], afterok: [], afternotok: [], afterany:
...
+    if script.copy_environment?
+      args += ["-env", (["all"] + env.keys).join(",")]
+    elsif script.copy_environment? == false
```
ericfranz: I know tests catch this, but could you add a short comment here that we are explicitly comparing to false because in the nil case we don't want to add any args?

ericfranz:
Are you aware of the ramifications of having both `-v` and `-V` appear as arguments to `qsub` in SGE and PBSPro?

MorganRodgers:
I have not yet tested them on live systems yet. That's one of the reasons why I am updating `ood-images` right now. Are you aware of any negative interactions between those flags / options?

ericfranz:
No I am not.

MorganRodgers:
I am going to test GridEngine and PBSPro tomorrow once I work through some Vagrant issues.

MorganRodgers:
GridEngine and PBSPro both behave as I would expect.

```bash
# Given a job.sh that echos its environment
A_ENV=value qsub job.sh                            # Does not output A_ENV
A_ENV=value qsub -V job.sh                         # Does output A_ENV
A_ENV=value qsub -v B_ENV=something_else job.sh    # Outputs B_ENV, but not A_ENV
A_ENV=value qsub -V -v B_ENV=something_else job.sh # Outputs A_ENV and B_ENV
```

PR #165, opened 2019-12-17
Add support for using Script#native to specify an override for the submit_host

### LINKED ISSUE #164: Support submission to specific nodes in the LinuxHostAdapter
```
One other question. Is it possible to have more than one submit_host?
And/or to select what particular ssh_host to use?

The reason for this question is that we have 8 interactive only nodes, named 
frisco[1-8].chpc.utah.edu, that don’t face any cluster and which we use for
interactive work. Each of these machines has different hardware specs, so, users
commonly request a specific machine (e.g. frisco8 that has more CPUs/memory
if they need them).
```

This could be supported by adding a specific SSH target (`submit_host_override`?) in the `Script#native`. The target node would need to be in the `ssh_hosts` so that job status polling would continue to work. If no specific target node is specified we would default to the `submit_host`.

MorganRodgers:
Fixes #164.

### DIFF
--- a/lib/ood_core/job/adapters/linux_host/launcher.rb
+++ b/lib/ood_core/job/adapters/linux_host/launcher.rb
@@ -57,7 +57,7 @@ class OodCore::Job::Adapters::LinuxHost::Launcher
   # @param hostname [#to_s] The hostname to submit the work to
   # @param script [OodCore::Job::Script] The script object defining the work
   def start_remote_session(script)
-    cmd = ssh_cmd(submit_host)
+    cmd = ssh_cmd((script.native && script.native['submit_host_override']) ? script.native['submit_host_override'] : submit_host)
 
     session_name = unique_session_name
     output = call(*cmd, stdin: wrapped_script(script, session_name))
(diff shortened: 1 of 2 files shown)


johrstrom:
seems fine to me, need to create a ticket on the docs to update.

lib/ood_core/job/adapters/linux_host/launcher.rb
```diff
@@ -57,7 +57,7 @@ def initialize(
   # @param hostname [#to_s] The hostname to submit the work to
   # @param script [OodCore::Job::Script] The script object defining the work
   def start_remote_session(script)
-    cmd = ssh_cmd(submit_host)
+    cmd = ssh_cmd((script.native && script.native['submit_host_override']) ? script.native['submit_host_override'] : submit_host)
```
ericfranz: Instead of `attr_reader :submit_host` and then this line we should have a custom getter for submit host:

```ruby
def submit_host(script = nil)
  if script && script.native && script.native['submit_host_override']
    script.native['submit_host_override']
  else
    submit_host
  end
end
```

then we keep `cmd = ssh_cmd(submit_host)`.

ericfranz: err we change that to `cmd = ssh_cmd(submit_host(script))`

PR #171, opened 2019-12-27
Support reading cluster configurations that are .yaml

treydock:


lib/ood_core/clusters.rb
```diff
@@ -27,10 +27,10 @@ def load_file(path)
             end
           end
         elsif config.directory?
-          Pathname.glob(config.join("*.yml")).select(&:file?).select(&:readable?).each do |p|
+          Pathname.glob([config.join("*.yml"), config.join("*.yaml")]).select(&:exist?).select(&:readable?).each do |p|
```
ericfranz: Why change `.select(&:file?)` to `.select(&:exist?)`?

treydock: In the rare case that someone uses symlinks.

MorganRodgers: `Pathname#file?` handles symlinks to files and directories transparently. This opens us up to the even less likely case that someone creates a directory named `something.yaml`.

treydock: Ah in that case I reverted back to using `file?`.

PR #173, opened 2020-01-06
Sanitize job names

### LINKED ISSUE #172: Add ability to sanitize job name in PBSPro and Grid Engine adapter
In some deployments of Grid Engine, special characters such as slash are not allowed in job names. This causes problems when submitting interactive apps, where we use slashes in the job name.

See https://github.com/OSC/ondemand/issues/341 and in particular https://github.com/OSC/ondemand/issues/341#issuecomment-567733351.

Suggested solution https://github.com/OSC/ondemand/issues/341#issuecomment-570267574

MorganRodgers:
Fixes #172.

### DIFF
--- a/lib/ood_core/job/adapters/pbspro.rb
+++ b/lib/ood_core/job/adapters/pbspro.rb
@@ -201,6 +201,7 @@ module OodCore
 
           @pbspro = o.fetch(:pbspro) { raise ArgumentError, "No pbspro object specified. Missing argument: pbspro" }
           @qstat_factor = o.fetch(:qstat_factor, 0.10).to_f
+          @sanitize_job_name = o.fetch(:sanitize_job_name, false)
         end
 
         # Submit a job with the attributes defined in the job template instance
@@ -237,7 +238,7 @@ module OodCore
           elsif script.email_on_terminated
             args += ["-m", "e"]
           end
-          args += ["-N", script.job_name] unless script.job_name.nil?
+          args += ["-N", job_name(script.job_name, @sanitize_job_name)] unless script.job_name.nil?
           args += ["-S", script.shell_path] unless script.shell_path.nil?
           # ignore input_path (not defined in PBS Pro)
           args += ["-o", script.output_path] unless script.output_path.nil?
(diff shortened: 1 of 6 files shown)


ericfranz:
Instead of a config option and force sanitizing every job submission... this might be a better alternative https://github.com/OSC/ood_core/issues/172#issuecomment-571255666

Then this method could be used instead of `parameterize.underscore`...

MorganRodgers:
I'm confused by the recommendations. For context the linked alternative:

> An `Adapter#sanitize_job_name(job_name)` method could be added whose default implementation just returns the job name in question. Minimally, that could be used in places where we submit jobs to first sanitize the job name.

I do not understand the utility of adding a super class implementation. Adding another method with a default implementation only extends the `Adapter` interface. We do still want a configuration option because presumably there are PBSPro and GE sites that do not have this problem, otherwise I would have expected that we would have heard about this sooner, and if a site is not experiencing the issue then we do not want/need to change OnDemand's behavior.

Just so I am clear your comment:

> Then this method could be used instead of `parameterize.underscore`...

By `this method` you are referencing the replacement of non-ASCII (or whatever the illegal characters set is) with underscores?

ericfranz:
If that method exists, then instead of https://github.com/OSC/ondemand/pull/364/files#diff-9ddf794da41002473c7abf3dc34e48d5R425

we could do something like

```ruby
def job_name
  [
    ENV["OOD_PORTAL"].to_s,
    ENV["RAILS_RELATIVE_URL_ROOT"].to_s.sub(/^\/[^\/]+\//, ""),
  ].reject(&:blank?).join("/").tap do |jb_nm|
    adapter.sanitize_job_name(jb_nm)
  end
end
```

ericfranz:
It is an alternative to the approach of providing a configuration flag that if enabled we always sanitize. But I guess that there are reasons we discussed offline where we might want to always sanitize every job name set...

ericfranz:
Opted for a simpler approach for now https://github.com/OSC/ood_core/pull/183 will revisit this more robust approach later.

Issue #178, opened 2020-02-06
Empty cluster config file causes crash on load


ericfranz:
This line crashes if the file exists but is empty: https://github.com/OSC/ood_core/blob/c1adab6a0e85671736b1484ebc5317eab28c6c00/lib/ood_core/clusters.rb#L32

ericfranz:
See also https://github.com/OSC/ood_core/issues/150

PR #181, opened 2020-03-05
Catch expections in parsing bad and empty cluster files

### LINKED ISSUE #150: Catch exceptions when reading bad cluster yml files

johrstrom:
This can catch Psych::SyntaxError while parsing a malformed cluster definition file and safely traverse empty cluster definitions.  This tries to fix #150 and #178. 

Most uses of Clusters filter off of `allow?` or `custom_allow?` and the InvalidCluster doesn't allow for anything so most (if not all) uses of clusters will filter these types out. 

Right now the only things you can really rely on are the `id` and `metadata.error_msg`.

### DIFF
--- a/lib/ood_core.rb
+++ b/lib/ood_core.rb
@@ -2,6 +2,7 @@ require "ood_core/version"
 require "ood_core/errors"
 require "ood_core/cluster"
 require "ood_core/clusters"
+require "ood_core/invalid_cluster"
 
 # The main namespace for ood_core
 module OodCore
--- a/lib/ood_core/cluster.rb
+++ b/lib/ood_core/cluster.rb
@@ -159,6 +159,12 @@ module OodCore
       }
     end
 
+    # This cluster is always valid
+    # @return true
+    def valid?
+      return true
+    end
+
     private
       # Build acl adapter objects from array
       def build_acls(ary)
--- a/lib/ood_core/cluster.rb
+++ b/lib/ood_core/cluster.rb
@@ -28,6 +28,10 @@ module OodCore
     # @return [Hash] the acls configuration
     attr_reader :acls_config
 
+    # The errors encountered with configuring this cluster
+    # @return Array<String> the errors
+    attr_reader :errors
+
     # @param cluster [#to_h] the cluster object
     # @option cluster [#to_sym] :id The cluster id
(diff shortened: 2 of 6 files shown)


johrstrom:
notes from discussion

* add a valid? api
*

johrstrom:
I added the `valid?` where clusters always return true and invalid_clusters always return false.

We had talked about adding an `error_msg` API too instead of holding the error message in metadata. Do you have more thoughts on that?  I only hesitated just as a reaction against expanding API further, but I'm happy to add it directly instead of burying it within metadata.

ericfranz:
One consideration is adding an API similar to ActiveModel validations i.e. https://api.rubyonrails.org/classes/ActiveModel/Errors.html. Not in the sense that in ActiveModel::Errors the error messages are tied to attributes. Rather, in the sense that `Cluster#errors` could be a hash or an array with one or more problems. It could be a simple as an array of strings, that by default is empty.

The reason is that you could have a valid cluster config, (valid yaml, minimum data required is there) but a section of it could be indented incorrectly and thus omitted (job section is indented incorrectly, so job management is not configured for the cluster).

Of course that type of validation and help is something that would be nice to do but we can save for the future. So if we extended the API at all that is what I would do.

ericfranz:
An example use of that would be a valid cluster config with "errors" would display using warning, and invalid cluster config would use error alert on dashboard.

johrstrom:
Done! I guess a hash would be better, then you could give severity? or we could add a `warnings` later to indicate errors are fatal and warnings, are well.. warnings. 

In any case, I've set it to an array of strings, if you want the hash instead just let me know.

PR #183, opened 2020-03-16
Add sanitize job name helper

ericfranz:
This addresses the base concerns of #172 

You can opt in by setting one or more illegal chars in env var OOD_JOB_NAME_ILLEGAL_CHARS. If you do this then these chars will be replaced with a dash when the helper method is used. The helper method can then be used in the dashboard app when submitting jobs.

This is the simplest and least disruptive change. It omits obvious enhancements like:

- using squeeze to remove consecutive dashes
- enabling the replacement char to be configured
- enabling dash to be specified as illegal
- enabling adapter or cluster specific modifications
- providing adapter specific defaults based on documentation of different schedulers

While all of these things are possible, these add more complexity and so this PR postpones that work till we get a new requirements.


johrstrom:
That seems OK, I think I still prefer ways that would work out of the box for everyone, but I suppose this a greater good situation.

ericfranz:
The only problem with this solution is those methods really don't belong there. Right now there isn't anything that changes from adapter to adapter in regards to this. There is only anticipation of a future need for this. Will sleep on it before merging.

ericfranz:
I added https://github.com/OSC/ondemand/issues/425 to revisit this in the future.

PR #185, opened 2020-04-15
ccq adapter

johrstrom:
The CCQ Adapter is for Cloudy Cluster sites.  ccq[sub,stat,del] are abstractions over q[sub,stat,del] for a given scheduler like Slurm or Torque.  It does not implement hold or release becuase there is
no ccqhold or ccqrel command.

Placing this initially in draft because it's only tested for Slurm CCQ clusters.  I'm testing Torque now, but am a little blocked on it and I doubt a whole lot could change even when I get unblocked.


johrstrom:
I'd like to get this into the 1.8 release if possible so that the CC team can start using it and integrate it with their install.

lib/ood_core/job/adapters/ccq.rb
```diff
@@ -0,0 +1,300 @@
...
+          # must at least have job_owner to filter by job_owner
+          attrs = Array.wrap(attrs) | [:job_owner] unless attrs.nil?
+
+          info_all(attrs: attrs).select { |info| owner.include? info.job_owner }
+        end
+
+        # Retrieve job info from the resource manager
+        # @param id [#to_s] the id of the job
+        # @return [Info] information describing submitted job
+        def info_all_each(attrs: nil)
```
ericfranz, line 96: info_all_each, info_where_owner both are just duplicate implementations of the base class so can be removed

lib/ood_core/job/adapters/ccq.rb
```diff
@@ -0,0 +1,300 @@
...
+          # columns that ccqstat gives.
+          info_from_ccqstat_extended(stat_output)
+        end
+
+        # Retrieve job status from resource manager
+        # @param id [#to_s] the id of the job
+        # @return [Status] status of job
+        # @see Adapter#status
+        def status(id)
+          info(id).status
```
ericfranz, line 125: If there is a `ccqstat` that returns less information than the info that might be preferred but that is a future optimization

lib/ood_core/job/adapters/ccq.rb
```diff
@@ -0,0 +1,300 @@
...
+          file.flush
+          file
+        end
+
+        def tmp_file_name
+          'ccq_ood_script_'
+        end
+
+        def parse_job_id_from_ccqsub(output)
+          match_data = /job id is: (?<job_id>\d+) you/.match(output)
```
ericfranz, line 230: This looks brittle in several respects.

1. will the job id always be 2 or more digits of an integer value (never starting with 1, 2, 3), regardless of whether Torque or Slurm is used?
2. will ccqsub always output the job id in between the words "job id is: " and " you"?

We had/have something similar to parse the job ID from LSF bsub output and it was problematic between different LSF versions. 

It might be worth requesting the CCQ developers to add to ccqsub support for a new argument that ensures only the job id is printed to stdout, instead of all of the human readable language around the job id

FWIW I did notice in the example text the only digits were either in the name of the host (slurm1) or the job id itself (123456). So perhaps the regex could be simpler if we knew more about it. Or maybe the regex used should actually be configurable - something that we can put in the cluster config, with a sane default - so at least our guess can be fixed without a code change if a situation arises.

lib/ood_core/job/adapters/ccq.rb
```diff
@@ -0,0 +1,300 @@
...
+          end
+
+          data_hash
+        end
+
+        def info_from_ccqstat(data)
+          infos = []
+
+          data.to_s.each_line do |line|
+            words = line.split(/\s/).reject(&:empty?)
```
ericfranz, line 273: Could the job name ever contain spaces? Can we specify the delimiter with ccqstat like we do with Slurm?

johrstrom: Good catch. They absolutely can contain spaces (I just tested this).  I think the formatting is a bit off anyhow (my username `johrstrom` turns into `johrs...`) so we may be able to work with them on that for correctness.

lib/ood_core/job/adapters/ccq.rb
johrstrom: So we spoke with the CCQ team and they indicated they'd at least keep this stable and backward compatible.  We're slightly lucky in that we can release atomically with them when they and it's cloud so folks will just replace instead of upgrade.

Either way I've made it configurable, but we're still looking for the named group `job_id` so there's a little extra in there in case someone sets the regex to look for `jobid` or similar, it'll at least throw an error that indicates the issue and not an uncaught exception.

lib/ood_core/job/adapters/ccq.rb
```diff
@@ -0,0 +1,300 @@
...
+          data_hash[:dispatch_time] = raw['startTime'].to_i
+          data_hash[:submission_time] = raw['dateSubmitted'].to_i
+          data_hash[:queue_name] = raw['criteriaPriority']
+
+          Info.new(data_hash)
+        end
+
+        # extended data is just lines of 'key: value' value, so parse
+        # it and stick it all in a hash.
+        def extended_data_to_hash(data)
```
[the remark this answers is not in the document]
johrstrom: Done! This is really slick.  The header I'm not worried about, it'll just get shoved into native if it exists. We extract specific items to populate Info, so worst case there's an odd thing in native data.

lib/ood_core/job/adapters/ccq.rb
johrstrom: There is, but it's unclear whether it'd actually be faster or not because while it is less data about any _one_ job, it is data for _all_ jobs (also `completed` state sticks around until deleted by a user).

lib/ood_core/job/adapters/ccq.rb
johrstrom: This is the only thing I can't solve immediately and quickly. It is well formatted from the header (see [this fixture](https://github.com/OSC/ood_core/blob/33088fdbe82844103ea4baac5a0b212d90285a95/spec/fixtures/output/ccq/ccqstat)) so we may be able to store the string index of the column. I'll reach out to the CCQ team about it.

ericfranz: We can merge this and open an issue to address this issue.

PR #188, opened 2020-05-05
lha bug fixes

johrstrom:
- commands are written in bash, so the ssh commands are now specifying bash
  to support users who have different login shells.
- -l given to pstree for long output becuase without it, it cuts off needed
  output.

### DIFF
--- a/lib/ood_core/job/adapters/linux_host/launcher.rb
+++ b/lib/ood_core/job/adapters/linux_host/launcher.rb
@@ -127,7 +127,7 @@ class OodCore::Job::Adapters::LinuxHost::Launcher
       [
         'ssh', '-t',
         '-o', 'BatchMode=yes',
-        "#{username}@#{destination_host}",
+        "#{username}@#{destination_host}"
       ].concat(cmd)
     else
       [
@@ -135,7 +135,7 @@ class OodCore::Job::Adapters::LinuxHost::Launcher
         '-o', 'BatchMode=yes',
         '-o', 'UserKnownHostsFile=/dev/null',
         '-o', 'StrictHostKeyChecking=no',
-        "#{username}@#{destination_host}",
+        "#{username}@#{destination_host}"
       ].concat(cmd)
     end
   end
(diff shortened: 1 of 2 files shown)


lib/ood_core/job/adapters/linux_host/launcher.rb
```diff
@@ -116,27 +116,27 @@ def call(cmd, *args, env: {}, stdin: "")
...
   # -o StrictHostKeyChecking=no (do no check the user's known hosts file)
-  def ssh_cmd(destination_host)
+  def ssh_cmd(destination_host, cmd)
     if strict_host_checking
       [
         'ssh', '-t',
         '-o', 'BatchMode=yes',
-        "#{username}@#{destination_host}"
-      ]
+        "#{username}@#{destination_host}",
```
ericfranz, line 130: since you are just adding concat, the trailing , is not necessary

lib/ood_core/job/adapters/linux_host/launcher.rb
```diff
@@ -116,27 +116,27 @@ def call(cmd, *args, env: {}, stdin: "")
...
+      ].concat(cmd)
     else
       [
         'ssh', '-t',
         '-o', 'BatchMode=yes',
         '-o', 'UserKnownHostsFile=/dev/null',
         '-o', 'StrictHostKeyChecking=no',
-        "#{username}@#{destination_host}"
-      ]
+        "#{username}@#{destination_host}",
```
ericfranz, line 138: since you are just adding concat, the trailing , is not necessary

lib/ood_core/job/adapters/linux_host/launcher.rb
```diff
@@ -116,27 +116,27 @@ def call(cmd, *args, env: {}, stdin: "")
     s.success? ? o : raise(Error, e)
   end
 
-  # The SSH invocation to send a command
+  # The SSH invocation to send the command you pass into cmd (Array<String>)
```
ericfranz, line 119: I had to read this sentence several times before I realized that "command you pass into cmd (Array<String>)" is referring to the positional argument `cmd`.

It might preferable to add the yardoc description for the positional arguments here, below the description i.e.

    @param destination_host ...
    @param cmd ...

Issue #109, opened 2018-11-27
SLURM: script.job_environment does not support --export=ALL ?


michaelkarlcoleman:
Not sure if this is missing doc or missing feature.  The `job_environment` YAML setting does not appear to support the `ALL` possibility, nor `NONE`, nor the export of a variable without specifying its value.  The `ALL` case is particularly important, as it looks like OOD is itself hard-coding a default of `NONE`.

A partial workaround that could be documented is to put `--export=ALL` in `native`.

From the `srun` man page:
```
       --export=<environment variables [ALL] | NONE>
              Identify which environment variables are propagated  to  the  launched
              application,  by  default  all  are  propagated.  Multiple environment
              variable names should be comma separated.  Environment variable  names
              may  be  specified to propagate the current value (e.g. "--export=EDI-
              TOR")  or  specific  values  may  be  exported  (e.g.   "--export=EDI-
              TOR=/bin/emacs").   In  these  two examples the environment propagated
              will only contain the variable "EDITOR"  and  nothing  else.   If  one
              desires  to  add  to  the environment instead of replacing it have the
              argument include ALL (e.g. "--export=EDITOR,ALL").  This  will  propa-
              gate  "EDITOR"  along with the current environment.  If one desires no
              environment variables be propagated use the argument NONE.  Regardless
              of  this setting, the appropriate "SLURM_*" task environment variables
              are always exported to the environment.
```

ericfranz:
@michaelkarlcoleman what does your job_environment yaml look like?

I would expect that you would be able to do this:

```yaml
script:
  job_environment:
    SBATCH_EXPORT: "ALL"
```

Are you doing this and it still not working?

I'm also curious what environment variables you want to propogate from the PUNs. I think the value of SBATCH_EXPORT being set to NONE by default was because when launching the PUNs sudo strips the environment, so the user's bashrc/bash_profile, for example, will not be sourced and the PUN environment will contain env vars only relevant to the web environment, not the compute environment. Perhaps these assumptions are wrong. Understanding your use case will be helpful when we update the documentation.

As for exporting specific environment variables instead of none, I guess with `job_environment` if `SBATCH_EXPORT` worked we could specify which env vars to set instead of `ALL`.

FWIW [this commit](https://github.com/OSC/ood_core/commit/4092ac17d9b6ae9618ad39c8a9e7c97f6775ef79) introduced this, though unfortunately the commit doesn't have a good comment so it isn't clear why.

[...]

PR #193, opened 2020-05-26
fix SLURM env setup

johrstrom:
fix SLURM environment setup such that --export=NONE is the default
(as it was before with SBATCH_EXPORT=NONE), job_environment adds
correctly to --export, and copy_environment correctly prefixes with ALL.

comment also left in the code as to why we default to NONE.

Fixes the issues raised in and created by #109 and #158.

### DIFF
--- a/lib/ood_core/job/adapters/slurm.rb
+++ b/lib/ood_core/job/adapters/slurm.rb
@@ -394,21 +394,7 @@ module OodCore
 
           # Set environment variables
           env = script.job_environment || {}
-          # we default to export NONE, but SLURM defaults to ALL.
-          # we do this bc SLURM setups a new environment loading /etc/profile (modules)
-          # where the PUN (the env sbatch uses) did not.
-          export = if !env.empty? && !script.copy_environment?
-                     env.keys.join(",")
-                   elsif !env.empty? && script.copy_environment?
-                     "ALL," + env.keys.join(",")
-                   elsif env.empty? && script.copy_environment?
-                     # only this option changes behaivor dramatically
-                     "ALL"
-                   else
-                     "NONE"
-                   end
-
-          args += ["--export", export]
+          args += ["--export", export_arg(env, script.copy_environment?)]
 
           # Set native options
           args += script.native if script.native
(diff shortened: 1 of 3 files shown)


lib/ood_core/job/adapters/slurm.rb
```diff
@@ -394,10 +394,21 @@ def submit(script, after: [], afterok: [], afternotok: [], afterany: [])
...
           # Set environment variables
           env = script.job_environment || {}
-          unless (script.job_environment.nil? || script.job_environment.empty?)
-            prefix = script.copy_environment? ? "ALL," : "NONE,"  # NONE if false or nil
-            args += ["--export", prefix + script.job_environment.keys.join(",")]
-          end
+          # we default to export NONE, but SLURM defaults to ALL.
+          # we do this bc SLURM setups a new environment loading /etc/profile (modules)
+          # where the PUN (the env sbatch uses) did not.
+          export = if !env.empty? && !script.copy_environment?
```
ericfranz, line 400: this block should be refactored to a method (private?), something like `def sbatch_export_environment_option(job_environment, copy_environment)` and then just `          args += ["--export", sbatch_export_environment_option(env, script.copy_environment?)]`

ericfranz:
By fixing it so we default to export NONE, how will that impact sites that already observed this problem and "fixed it". Will their "fixes" still work, or will they negatively affected by the switch again?

spec/job/adapters/slurm_spec.rb
```diff
@@ -41,6 +41,10 @@ def build_script(opts = {})
       )
     end
 
+    def base_args(arr)
+      arr + ["--export", "NONE"]
+    end
```
ericfranz, line 46: Is there a better name than base_args? Just by going by the name of the method, it isn't clear why every test wraps the args to test against with base_args except for the test `context "with :job_environment" do` where it is omitted and `args: ["--export", "key"]` is used. It looks like the fact that it isn't used in the latter test is a bug but of course adding it would break the test.

I don't know the answer besides a larger refactor that would change/simplify how we test these things.

johrstrom:
I can test it to be doubly sure, but the 'fix' was to source files under `/etc/profile.d`.  Given that `--export=NONE` seems to already do this, sourcing the file(s) a second time would not seem to change anything.

Which is to say, sites with the 'fix' wouldn't be impacted other than having a little bit of tech debt to clean up.  But I'll confirm.

spec/job/adapters/slurm_spec.rb
ericfranz: It would be better to get rid of base_args and inline with the rest of the tests. Since we arbitrarily add base args to the tests, it isn't the right abstraction. Better to have duplication than the wrong abstraction. A better architecture for the adapter code in the future will address the problems with the tests.

So

```diff
    context "with :queue_name" do
      before { adapter.submit(build_script(queue_name: "queue")) }

-     it { expect(slurm).to have_received(:submit_string).with(content, args: ["-p", "queue"], env: {}) }
+     it { expect(slurm).to have_received(:submit_string).with(content, args: ["-p", "queue", "--export", "NONE"], env: {}) }
    end
```

johrstrom:
I've been double checking this morning and I think we're OK with the 2nd sourcing of files. The site I was working on even had it in an if block (if `-z $LMOD_VERSION`), but again, sourcing a profile file a second time shouldn't cause any problem.

PR #198, opened 2020-06-09
Attributes from qstat command now found under native hash

matthu017:
Added cpu/mem_usage attributes following gridengine/qstat.xsd

### DIFF
--- a/lib/ood_core/job/adapters/sge/qstat_xml_j_r_listener.rb
+++ b/lib/ood_core/job/adapters/sge/qstat_xml_j_r_listener.rb
@@ -28,10 +28,13 @@ class QstatXmlJRListener
       :tasks => [],
       :status => :queued,
       :procs => 1,
-      :native => {}  # TODO: improve native attribute reporting
+      :native => {
+        :ST_name = ''
+      }  # TODO: improve native attribute reporting
     }
     @current_text = nil
     @current_request = nil
+    @stdout_PN_path = false
 
     @processing_job_array_spec = false
     @adding_slots = false
@@ -42,6 +45,7 @@ class QstatXmlJRListener
       step: 1,  # Step can have a default of 1
     }
     @running_tasks = []
+    @native_tags = ['JB_job_number', 'JB_job_name', 'JB_version', 'JB_project', 'JB_exec_file', 'JB_script_file', 'JB_script_size', 'JB_submission_time', 'JB_execution_time', 'JB_deadline', 'JB_owner', 'JB_uid', 'JB_group', 'JB_gid', 'JB_account', 'JB_cwd', 'JB_notify', 'JB_type', 'JB_reserve', 'JB_priority', 'JB_jobshare', 'JB_verify', 'JB_checkpoint_attr', 'JB_checkpoint_interval', 'JB_restart']
   end
 
   def tag_start(name, attrs)
@@ -50,13 +54,15 @@ class QstatXmlJRListener
       toggle_processing_array_spec
     when 'JB_pe_range'
       toggle_adding_slots
+    when 'JB_stdout_path_list'
+      @stdout_PN_path = true
     end
   end
 
   def tag_end(name)
-    #Add this to native hash if there is a text node
-    if (!@current_text.strip.empty?)
-      @current_job[:native][name] = @current_text
+    #Add to native hash if in native_tags
+    if (@native_tags.include?(name))
+      @current_job[:native][:"#{name}"] = @current_text
     end
 
     case name
@@ -97,6 +103,10 @@ class QstatXmlJRListener
       toggle_processing_array_spec
     when 'JB_pe_range'
       toggle_adding_slots
+    when 'PN_path' 
+      end_PN_path
+    when 'ST_name'
+      end_ST_name
     end
   end
 
@@ -176,6 +186,17 @@ class QstatXmlJRListener
   # Used to record a running Job Array task
   def end_JAT_task_number
     @running_tasks << @current_text
+  end
+
+  def end_PN_path
+    if (@stdout_PN_path)
+      @stdout_PN_path = false
+      @current_job[:native][:PN_path] = @current_text
+    end
+  end
+
+  def end_ST_name
+    @parsed_job[:native][:ST_name] = @parsed_job[:native][:ST_name] + @current_text + ' '
   end
 
   def set_job_array_piece(key)
--- a/lib/ood_core/job/adapters/sge/qstat_xml_j_r_listener.rb
+++ b/lib/ood_core/job/adapters/sge/qstat_xml_j_r_listener.rb
@@ -54,6 +54,11 @@ class QstatXmlJRListener
   end
 
   def tag_end(name)
+    #Add this to native hash if there is a text node
+    if (!@current_text.strip.empty?)
+      @current_job[:native][name] = @current_text
+    end
+
     case name
     when 'JB_ja_tasks'
       end_JB_ja_tasks
@@ -92,10 +97,6 @@ class QstatXmlJRListener
       toggle_processing_array_spec
     when 'JB_pe_range'
       toggle_adding_slots
-    when 'cpu_usage'
-      end_cpu_usage
-    when 'mem_usage'
-      end_mem_usage
     end
   end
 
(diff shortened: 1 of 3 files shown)


matthu017:
The XML schema for qstat can be found at the link below: https://github.com/gridengine/gridengine/blob/master/source/dist/util/resources/schemas/qstat/qstat.xsd
The `vagrant-with-gridengine` under OSC's `ood-images` repo did not have many of these XML tags but it should probably work for other HPC centers using SGE. As seen in the discourse below, these tags are present under other production instances: 
https://discourse.osc.edu/t/remote-desktop-not-reflecting-correct-number-of-cores/640

matthu017:
Need to change unit testing to read values from `native` hash.
Change dispatch/submission time from UNIX time to a human readable format

ericfranz:
The conversion of these hashes into Info objects is done https://github.com/OSC/ood_core/blob/63187f8874e62b752a73e39263f539ce0241495a/lib/ood_core/job/adapters/sge/batch.rb#L68-L72.

The idea is the primary purpose of the info and info_all methods on the adapters is to return an array of objects that have the same interface, regardless of the adapter.

So id, job_owner, job_name, accounting_id, status, procs, queue_name, submission_time, dispatch_time, wallclock_limit etc. are all defined https://github.com/OSC/ood_core/blob/63187f8874e62b752a73e39263f539ce0241495a/lib/ood_core/job/info.rb.

The original idea for native was when building the Slurm and Torque adapters - the Info interface was small, and the qstat/squeue calls returned a lot more information than was in our interface, and Jeremy wanted to capture the original full data somewhere in that object for access.

For example, the actual state in Torque might be Exiting or Transitioning but these exact states do not exist in every adapter. So the common interface to job status is OodCore::Job::Status which Info returns. But native would provide a way to get at the raw data produced by Torque, which would include the original T or E for the status. The native data would also use the original key names. So for example, Info#queue_name in Slurm native attribute might have the key name "partition".

Here is an example of creating an Info object using a subset of the data from torque, and then setting native to the raw data at the end: https://github.com/OSC/ood_core/blob/63187f8874e62b752a73e39263f539ce0241495a/lib/ood_core/job/adapters/pbspro.rb#L443-L459

For the SGE adapter, when parsing the XML we build a hash with the keys matching the info object, so initializing a new Info object is made easier - but of course the native attribute is empty.

There are several approaches.

1. Build the full native raw data from the XML, alongside the rest. So instead of replacing like this PR currently does:

    ```diff
      def end_JB_job_number
    -    @parsed_job[:id] = @current_text
    +    @parsed_job[:native][:id] = @current_text
      end
    ```

    it is done in addition:

    ```diff
      def end_JB_job_number
         @parsed_job[:id] = @current_text
    +    @parsed_job[:native][:job_number] = @current_text
      end
    ```

    or

    ```diff
      def end_JB_job_number
         @parsed_job[:id] = @current_text
    +    @parsed_job[:native][:JB_job_number] = @current_text
      end
    ```


    Notice that native key is "job_number" or "JB_job_number" not "id" because in SGE parlance it is called a job number, and that is what you see in SGE documentation and in the XSD schema https://github.com/gridengine/gridengine/blob/6a5407d56c85b39290ac2488fb6dec1a4404a974/source/dist/util/resources/schemas/qstat/qstat.xsd#L113

    With this approach, you would add code to capture other tags if you run across them, adding them to native. Examples: slots, tasks, master, tickets, deadline, priority, etc. These are obviously all candidates https://github.com/gridengine/gridengine/blob/6a5407d56c85b39290ac2488fb6dec1a4404a974/source/dist/util/resources/schemas/qstat/qstat.xsd#L113-L165. You could probably just use the short form in the XML if they match what a user might see when using qstat from the command line, or even the full XML tag name such as JB_job_number, JAT_prio, etc. 

There is likely a place in the sax parser to add every tag name and value to native - then it is just a change in one location.

ericfranz:
In fact if you can just add a snippet to capture all tags and put the text contents in the native hash, in the tag end method. But that might be problematic for cases where we have hierarchy of tags; or we might still want to cherry pick tags whose values we want to display to the user.

The XSD file above does not contain all the possible values different versions of SGE provide - just looking at the Discourse topic you mentioned, there are a bunch of tags that are not represented.

If we have the access to the output file path, that can be used for the links to the job directory in the progressive disclosure on the Active Jobs app (Files and Shell links).

matthu017:
I'm not seeing anything regarding output file path in the tags. Is there another way to get the path?

matthu017:
[NewAttributesForSGE.txt](https://github.com/OSC/ood_core/files/4755097/NewAttributesForSGE.txt)
This file details potential attributes we could support in the ood_core gem for grid engine

To address the differences between other versions of grid engine one could specify a .yaml or .json file under `etc/ood/config/apps/activejobs`. 
This would allow users to specify their own key-value pairs for attributes that show up in their `qstat -j -r -xml` in addition to the ones already supported by the gem.
``` attributes.json
JB_job_number: Job Number   //Or whatever key you want to show up in active jobs
```

lib/ood_core/job/adapters/sge/qstat_xml_r_listener.rb
```diff
@@ -24,21 +24,32 @@ def initialize
...
-      :native => {}  # TODO: improve native reporting
+      :native => {
+        :ST_name => ''
+      }  # TODO: improve native reporting
     }
     @current_text = nil
+    @stdout_PN_path = false
 
     @current_request = nil
+    @native_tags = ['JB_job_number', 'JB_job_name', 'JB_version', 'JB_project', 'JB_exec_file', 'JB_script_file', 'JB_script_size', 'JB_submission_time', 'JB_execution_time', 'JB_deadline', 'JB_owner', 'JB_uid', 'JB_group', 'JB_gid', 'JB_account', 'JB_cwd', 'JB_notify', 'JB_type', 'JB_reserve', 'JB_priority', 'JB_jobshare', 'JB_verify', 'JB_checkpoint_attr', 'JB_checkpoint_interval', 'JB_restart']
```
johrstrom, line 35: Some of these are already accounted for, like `JB_job_number`, `JB_job_name`. But even more so like `JB_submission_time` because in my PR we can't just extract it, we need to post process it.

johrstrom: I see, I guess it's just additive in native. It'll add to native, but _also_ add it to the common script attribute.

lib/ood_core/job/adapters/sge/qstat_xml_j_r_listener.rb
```diff
@@ -153,6 +168,17 @@ def end_JAT_task_number
...
+  def end_PN_path
+    if (@stdout_PN_path)
+      @stdout_PN_path = false
+      @parsed_job[:native][:PN_path] = @current_text
+    end
+  end
+
+  def end_ST_name
+    @parsed_job[:native][:ST_name] = @parsed_job[:native][:ST_name] + @current_text + ' '
+  end
```
johrstrom, line 180: What's `ST_name` supposed to for us, the output file?  If you see the [UGE xml in another PR](https://github.com/OSC/ood_core/blob/c2b64e6e39c300931530983afebd47f10b4bec25/spec/job/adapters/sge/output_examples/uge_qstat_jr.xml), it looks like it's an element for the submission command?  I can't find an example of it in our current test output files.

Either way, there's no test case for it, and I'm not sure if it would be end up being what you expect. (I imagine parsers take sets of ST_name, ST_pos and ST_access_specifier to rebuild the command).

johrstrom: OK I see this is additive, continually appending.

matthu017: The discourse page mentioned uses ST_name and it seems like the submission_args which is something shown in active jobs.

lib/ood_core/job/adapters/sge/qstat_xml_j_r_listener.rb
```diff
@@ -28,10 +28,13 @@ def initialize
       :tasks => [],
       :status => :queued,
       :procs => 1,
-      :native => {}  # TODO: improve native attribute reporting
+      :native => {
+        :ST_name => ''
+      }  # TODO: improve native attribute reporting
```
johrstrom, line 33: I think we can remove this TODO now.

johrstrom:
I think we're good with this now.  Though, I'd ask that you rebase to master and use the uge test file to make a testcase for `native.ST_name`.  Even if we're not going to use it, we'll have some coverage for when/if we do.

PR #200, opened 2020-06-10
checks for similarity between ssh_host and hostname -A

matthu017:
Not sure if this is the right way to raise an exception. Usually seen in code as `rescue Error => e` 
May be inefficient because it turns all output into potential `hostnames`

### DIFF
--- a/lib/ood_core/job/adapters/linux_host/launcher.rb
+++ b/lib/ood_core/job/adapters/linux_host/launcher.rb
@@ -64,12 +64,8 @@ class OodCore::Job::Adapters::LinuxHost::Launcher
     hostnames = output.strip.split #Will capture all other data too
     ssh_host_array = @ssh_hosts.to_a #is a Set seen in linux_host.rb
     hostname = (hostnames & ssh_host_array).first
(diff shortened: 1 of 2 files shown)


lib/ood_core/job/adapters/linux_host/launcher.rb
```diff
@@ -61,7 +61,10 @@ def start_remote_session(script)
 
     session_name = unique_session_name
     output = call(*cmd, stdin: wrapped_script(script, session_name))
-    hostname = output.strip
+    hostnames = output.strip.split #Will capture all other data too
+    ssh_host_array = @ssh_hosts.to_a #is a Set seen in linux_host.rb
+    hostname = (hostnames & ssh_host_array).first
+    raise Error, "The specified host is not in the list of ssh hosts configured for this cluster. The ssh hosts configured are #{ssh_host_array.inspect}. The specified host for this job is determined by running 'hostname -A' on the target host and this output must match one of the specified ssh hosts" unless !hostname.nil?
```
ericfranz, line 67: We need to do this in the bash script so that the command being executed via ssh fails - and aborts with an error, instead of continuing to star the processes.

ericfranz:
Oh sorry this is a duplicate of https://github.com/OSC/ood_core/pull/201. Closing this.

Issue #190, opened 2020-05-14
host_permitted? error difficult to debug


ericfranz:
We have a `host_permitted?` method that raises an exception if the metadata of a Linux Host adapter job contains the hostname of a host that is not in the list of allowed hosts.

This can occur in the common case where the `hostname` command on a target host does not return the FQDN but the FQDN is used in the cluster config for the ssh hosts to check the status of. The result is something like this: metadata for a job says "owens" is the host to check the status of, but the only host in ssh hosts for the cluster using the linux host adapter is "owens.hpc.osc.edu" so the error raised is "Requested destination host (owens) not permitted". This is written to the log files, and in the user interface it just shows the job as being in a bad state.

There are two things that need to change.

1. A better error message is needed than "Requested destination host (owens) not permitted". Something that explains the problem and suggests the solution. "The specified host 'owens-login01' is not in the list of ssh hosts configured for this cluster. The ssh hosts configured are 'owens-login01.hpc.osc.edu', 'owens-login02.hpc.osc.edu', 'owens-login03.hpc.osc.edu'. The specified host for this job is determined by running `hostname` on the target host. The output of `hostname` must match one of the specified ssh hosts."

2. This error message should appear in the batch connect panel when it is in a bad state, instead of being hidden in the log file.

ericfranz:
Note: https://github.com/OSC/ood_core/issues/191 will fix the issue with hostname in a more appropriate way. However, the fixes above are relevant for any time that getting the status of a job causes a problem.

matthu017:
Is this solved by PR: https://github.com/OSC/ood_core/pull/201? Or does this require a separate solution?

johrstrom:
> Is this solved by PR: https://github.com/OSC/ood_core/pull/201?

Yes this was fixed in 201.

PR #201, opened 2020-06-10
Process hostnames and ssh_host similarities in bash

matthu017:
Not sure how to write to STDERR so will have to be changed later. Alternative to processing in ruby.

### DIFF
--- a/lib/ood_core/job/adapters/linux_host/launcher.rb
+++ b/lib/ood_core/job/adapters/linux_host/launcher.rb
@@ -60,7 +60,11 @@ class OodCore::Job::Adapters::LinuxHost::Launcher
     cmd = ssh_cmd(submit_host(script), ['/usr/bin/env', 'bash'])
 
     session_name = unique_session_name
-    output = call(*cmd, stdin: wrapped_script(script, session_name))
+    ssh_host_delimited = ssh_host.to_a.join(':')
+    environment = {"ssh_host" => ssh_host_delimited}
+
+    output = call(*cmd, stdin: wrapped_script(script, session_name), env: environment)
+    
     hostname = output.strip
 
     "#{session_name}@#{hostname}"
--- a/lib/ood_core/job/adapters/linux_host/launcher.rb
+++ b/lib/ood_core/job/adapters/linux_host/launcher.rb
@@ -60,7 +60,7 @@ class OodCore::Job::Adapters::LinuxHost::Launcher
     cmd = ssh_cmd(submit_host(script), ['/usr/bin/env', 'bash'])
 
     session_name = unique_session_name
-    ssh_host_delimited = ssh_host.to_a.join(':')
+    ssh_host_delimited = ssh_hosts.to_a.join(':')
     environment = {"ssh_host" => ssh_host_delimited}
 
     output = call(*cmd, stdin: wrapped_script(script, session_name), env: environment)
--- a/lib/ood_core/job/adapters/linux_host/launcher.rb
+++ b/lib/ood_core/job/adapters/linux_host/launcher.rb
@@ -60,10 +60,7 @@ class OodCore::Job::Adapters::LinuxHost::Launcher
     cmd = ssh_cmd(submit_host(script), ['/usr/bin/env', 'bash'])
 
     session_name = unique_session_name
-    environment = {"SSH_HOSTS" => ssh_hosts.to_a.join(' ')}
-
-    output = call(*cmd, stdin: wrapped_script(script, session_name), env: environment)
-    
+    output = call(*cmd, stdin: wrapped_script(script, session_name))
     hostname = output.strip
 
     "#{session_name}@#{hostname}"
@@ -177,6 +174,7 @@ class OodCore::Job::Adapters::LinuxHost::Launcher
         'session_name' => session_name,
         'singularity_bin' => singularity_bin,
         'singularity_image' => singularity_image(script.native),
+        'ssh_hosts' => ssh_hosts,
         'tmux_bin' => tmux_bin,
       }.each{
         |key, value| bnd.local_variable_set(key, value)
--- a/lib/ood_core/job/adapters/linux_host/launcher.rb
+++ b/lib/ood_core/job/adapters/linux_host/launcher.rb
@@ -60,8 +60,7 @@ class OodCore::Job::Adapters::LinuxHost::Launcher
     cmd = ssh_cmd(submit_host(script), ['/usr/bin/env', 'bash'])
 
     session_name = unique_session_name
-    ssh_host_delimited = ssh_hosts.to_a.join(':')
-    environment = {"ssh_host" => ssh_host_delimited}
+    environment = {"SSH_HOSTS" => ssh_hosts.to_a.join(' ')}
 
     output = call(*cmd, stdin: wrapped_script(script, session_name), env: environment)
     
--- a/lib/ood_core/job/adapters/linux_host/templates/script_wrapper.erb.sh
+++ b/lib/ood_core/job/adapters/linux_host/templates/script_wrapper.erb.sh
@@ -1,5 +1,19 @@
 #!/bin/bash
-hostname
+hostnames=$(hostname -A)
+array_ssh_host=(${ssh_host//:/ }) #replace all : with " "
+hostnameFound=false
+for host in $array_ssh_host
+do
+    if [[ " ${hostnames[@]} " =~ " ${host} " ]]; then
+        hostnameFound=true
+        hostname=$host
+    fi
+done
+
+if [ ! $hostnameFound ]; then
+    echo "ERROR: The specified host is not in the list of ssh hosts configured for this cluster [${array_ssh_host[@]}]. The specified hosts determined by running 'hostname -A' [${hostnames[@]}] on the target host and must match one of the configured ssh hosts"
+    exit 1
+fi
 
 # Put the script into a temp file on localhost
 <% if debug %>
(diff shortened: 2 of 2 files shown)


lib/ood_core/job/adapters/linux_host/launcher.rb
```diff
@@ -60,7 +60,11 @@ def start_remote_session(script)
     cmd = ssh_cmd(submit_host(script), ['/usr/bin/env', 'bash'])
 
     session_name = unique_session_name
-    output = call(*cmd, stdin: wrapped_script(script, session_name))
+    ssh_host_delimited = ssh_hosts.to_a.join(':')
+    environment = {"ssh_host" => ssh_host_delimited}
```
ericfranz, line 64: When setting env vars here lets use the convention of uppercase:

```
environment = {"SSH_HOSTS" => ssh_hosts.to_a.join(':')}
```

It is also okay to have this be one line since `ssh_hosts.to_a.join(':')` is very readable as is.

ericfranz: Also is there a problem with space delimiting the value? That way you could omit the line in the bash script where you substitute `:` for `" "`

matthu017: Yes on second thought, Open3.capture3 supports spaces in their env variables (unlike FFI)

ericfranz:
@matthu017 go ahead and squash this PR into 1 commit then have @johrstrom review it

lib/ood_core/job/adapters/linux_host/templates/script_wrapper.erb.sh
```diff
@@ -1,5 +1,20 @@
...
+    if [[ " ${hostnames[@]} " =~ " ${host} " ]]; then
+        hostnameFound=true
+        hostname=$host
+    fi
+done
+
+if [ ! $hostnameFound ]; then
+    echo >&2 "ERROR: The specified host is not in the list of ssh hosts configured in this cluster config [${SSH_HOSTS[@]}]. The specified hosts determined by running 'hostname -A' [${hostnames[@]}] on the target host must match one of the configured ssh hosts" 
+    exit 1
+fi
```
johrstrom, line 15: This if statement doesn't work. I mean it never evaluates to false. 

I think if you get rid of the boolean and just check to see if `hostname` ever got set is a safer test. Bash/shells can be funny about booleans so I usually just write a string and is it's existence as the boolean value.
```bash
if [[ -n "${hostnameFound+x}" ]]; then
    # ...
fi
```

Also, since it's already in an erb, we should probably have a translatable error message.

lib/ood_core/job/adapters/linux_host/templates/script_wrapper.erb.sh
```diff
@@ -1,5 +1,20 @@
...
-hostname
+hostnames=$(hostname -A)
+hostnameFound=false
+for host in $SSH_HOSTS
+do
+    if [[ " ${hostnames[@]} " =~ " ${host} " ]]; then
+        hostnameFound=true
+        hostname=$host
+    fi
+done
```
johrstrom, line 10: I don't believe `SSH_HOSTS` is being set on the destination. I just got `[]` and I believe it's because it's evaluated on the other side, and we're not passing it through ssh with `SendEnv`, which may be impossible as sshd configures to allow only certain variables through `AcceptEnv`. 

I can't see what we accept because of permissions, but the point may be moot as you could just write out hosts as an environment variable inlined in the script, so you don't have to pass through ssh. 

Something like this, writing the ruby array out to a bash array:
`SSH_HOSTS=(<%= hosts.join(' ').to_s =>)`

ericfranz: Oh yes, this is quite different than the other adapters. So that also means that if `OodCore::Job::Script#job_environment` has values set, we are not setting these when we start the processes on the remote host?

johrstrom: It looks like we export them inline in the script, so we're not passing them through ssh.

This bit here gets written out as just `<%= environment %>`. 
https://github.com/OSC/ood_core/blob/9eaf3cfe937657104f7d16b4bf2452be18cbdebc/lib/ood_core/job/adapters/linux_host/launcher.rb#L184-L193

lib/ood_core/job/adapters/linux_host/templates/script_wrapper.erb.sh
matthu017: What do you mean by 'translatable error message'?

johrstrom: Now that I actually say it, I don't think ood_core deals with il8n. Rails does, which this is not, so you can disregard. 

But what I'd meant was Rails' I18n library so that messages can default to english but we'd have the ability to translate it to another language later through configuration.

lib/ood_core/job/adapters/linux_host/templates/script_wrapper.erb.sh
```diff
@@ -1,5 +1,19 @@
 #!/bin/bash
-hostname
+SSH_HOSTS=(<%= ssh_hosts.join(' ').to_s %>)
+hostnames=`hostname -A`
+for host in $SSH_HOSTS
```
johrstrom, line 4: `$SSH_HOSTS` needs to be `${SSH_HOSTS[@]}` to loop through all the elements in the array. As is it's just the first.

lib/ood_core/job/adapters/linux_host/templates/script_wrapper.erb.sh
```diff
@@ -1,5 +1,19 @@
...
+do
+    if [[ " ${hostnames[@]} " =~ " ${host} " ]]; then
+        hostname=$host
+    fi
+done
+
+if [ -z "$hostname" ]; then
+    echo >&2 "ERROR: The specified host is not in the list of ssh hosts configured in this cluster config [${SSH_HOSTS[@]}]. The specified hosts determined by running 'hostname -A' [${hostnames[@]}] on the target host must match one of the configured ssh hosts"
```
johrstrom, line 12: I wonder if this can't be worded a little simpler? Something like `ERROR: Can't start a job on '$(hostname -A)' because it's not like any hostname configured in ssh_hosts [${SSH_HOSTS[@]}]`.

lib/ood_core/job/adapters/linux_host/templates/script_wrapper.erb.sh
matthu017: Both $SSH_HOSTS AND ${SSH_HOSTS[@]} work as intended

lib/ood_core/job/adapters/linux_host/templates/script_wrapper.erb.sh
matthu017: It definitely can be. Not sure whether this issue is being combined with https://github.com/OSC/ood_core/issues/190. The error message was taken directly from that issue.

lib/ood_core/job/adapters/linux_host/templates/script_wrapper.erb.sh
johrstrom: I'm not sure I follow what you mean by "work as intended". I meant it needs changed this `for element in array` statement.

See the difference these one liners
` ARR=(one two three); for ele in $ARR; do echo $ele; done`
`ARR=(one two three); for ele in ${ARR[@]}; do echo $ele; done`

You'll see the first only loops through the first element, while the second loops through all three, because the second expands the array.

matthu017:
```#!bin/bash

SSHHOSTS='foo bar gee'
for host in $SSHHOSTS
do
        echo $host
done

for host in ${SSHHOSTS[@]}
do
        echo $host
done
```
The difference is my array is a string with multiple values

lib/ood_core/job/adapters/linux_host/templates/script_wrapper.erb.sh
johrstrom: It seems like this should fix #190 because it'll trap the most common case for this adapter being in a bad state.

Yea I think the key here is "Something that explains the problem and suggests the solution". I would add "in plain words". 

How about this:
`ERROR: Can't start a job on '$(hostname -A)' because it's not like any hostname configured in ssh_hosts [${SSH_HOSTS[@]}]. The output of 'hostname -A' must match an entry in ssh_hosts from the cluster configuration.`

Also, you can add newlines in there too. As 1 line you have to scroll pretty far to the right.

lib/ood_core/job/adapters/linux_host/templates/script_wrapper.erb.sh
matthu017: the erb statement ssh_hosts.join(' ').to_s will result in a string like 'foo bar gee'

johrstrom: Oh I see. What gets written though is `SSH_HOSTS=(foo bar gee)` which is a bash array.

One of the two has to change, because it doesn't work as is.

lib/ood_core/job/adapters/linux_host/templates/script_wrapper.erb.sh
```diff
@@ -1,5 +1,19 @@
...
+hostnames=`hostname -A`
+for host in ${SSH_HOSTS[@]}
+do
+    if [[ " ${hostnames[@]} " =~ " ${host} " ]]; then
+        hostname=$host
+    fi
+done
+
+if [ -z "$hostname" ]; then
+    printf >&2 "ERROR: Can't start job on [${hostnames[@]}] because it does not match any hostname configured \nin ssh_hosts [${SSH_HOSTS[@]}]. The output of 'hostname -A' must match an entry in ssh_hosts \nfrom the cluster configuration."
```
matthu017, line 12: can alternatively use `echo -e`

PR #202, opened 2020-06-15
Fix spaces for envvars in submit.yml.erb

matthu017:
Used `Shellwords.escape`

### DIFF
--- a/lib/ood_core/job/adapters/torque.rb
+++ b/lib/ood_core/job/adapters/torque.rb
@@ -1,5 +1,6 @@
 require "ood_core/refinements/hash_extensions"
 require "ood_core/job/adapters/helper"
+require 'shellwords'
 
 module OodCore
   module Job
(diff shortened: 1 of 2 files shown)


matthu017:
A more permanent fix would be to change script.native hash into array format but it seems like torque is the only one would need this (As it is the only one that uses `send()` instead of `Open3.capture3()` and only when script.native is a hash)

Open3.capture3 correctly submits the envvars (no shell escape needed)

lib/ood_core/job/adapters/torque.rb
```diff
@@ -129,6 +130,9 @@ def submit(script, after: [], afterok: [], afternotok: [], afterany: [])
               envvars.merge!   script.native.fetch(:envvars, {})
             end
 
+            # Destructively change envvars to shellescape values
+            envvars.each { |k, v| envvars[k] = Shellwords.escape(v)}
```
ericfranz, line 134: Use `Hash#transform_values!` instead:

    envvars.transform_values! { |v| Shellwords.escape(v) }

ericfranz:
As for changing the torque adapter to use only array for native attribute, this adapter supports both hash and array and you may have seen the comment at the top:

>          # FIXME: Remove the Hash option once all Interactive Apps are
>          # converted to Array format

It is something we could do for 2.0 if we wanted to deprecated this, but we would need to double check and ensure all our apps use array for the native option and then some how log a deprecation warning when using the hash approach.

Issue #170, opened 2019-12-20
Make the "ssh wrapper" pattern a first class citizen


ericfranz:
Like with the linux host adapter, in the job config if you specify `submit_host: "some_other_host"` then when executing the command to submit the batch job, each adapter would execute that command, its arguments, etc. on another host via ssh. This is the main scope of this feature.

A second feature would be to add another config `use_submit_host_for_all_commands: true` (a better name?), every adapter that doesn't use the C API should execute all of the scheduler commands on the specified submit host (qstat, qdel, qhold, etc.). This is problematic, however, for adapters that use the C API via FFI for anything (SGE, Torque).

Currently, the only way to use this "ssh wrapper" pattern is for a site to provide their own ssh wrapper scripts and use bin_overrides. This is onerous enough that it is not used by default, and we don't even document well how to do it.

ericfranz:
This would not work for sge or the torque adapters since they both make use of the C libraries. But this would work for Slurm and LSF.

MorganRodgers:
This should be doable in PBSPro as well, it would also work for SGE if the site can accept less nuanced job states by not using DRMAA.

ericfranz:
We could support two options: ssh for just job submission, and ssh for all commands

PR #204, opened 2020-06-19
Submit commands on another host via ssh

matthu017:
Made Adapters::Helper.ssh_wrap to return commands wrapped in ssh


lib/ood_core/job/adapters/helper.rb
```diff
@@ -12,6 +12,17 @@ class Helper
...
           bin_overrides.fetch(cmd.to_s) { Pathname.new(bin_default.to_s).join(cmd.to_s).to_s }
         end
+        
+        # Gets a command that submits command on another host via ssh
+        # @param cmd [String] the desired command to execute on another host
+        # @param submit_host [String] where to submit the command
+        # @return [String] command wrapped in ssh if submit_host is present
+        def self.ssh_wrap(cmd, submit_host)
+          if submit_host.empty?
+            return cmd
```
ericfranz, line 22: In Ruby it is typical for guard clauses like this to be a single line with a blank line after separating it from the rest of the code. So

```
return cmd if submit_host.empty?

"ssh #{submit_host} \"#{cmd}\""
```

lib/ood_core/job/adapters/helper.rb
```diff
@@ -12,6 +12,17 @@ class Helper
...
+          end
+          "ssh #{submit_host} \"#{cmd}\""
```
ericfranz, line 24: Check the implementation of this for the Linux host adapter: https://github.com/OSC/ood_core/blob/63187f8874e62b752a73e39263f539ce0241495a/lib/ood_core/job/adapters/linux_host/launcher.rb#L119-L145

In particular the known hosts issue and the `'-o', 'BatchMode=yes'` argument.

Note I'm going to open a another issue to see if we can handle known_hosts in a better way.

lib/ood_core/job/adapters/pbspro.rb
```diff
@@ -9,18 +9,18 @@ class Factory
...
-      # @option config [Object] :host (nil) The batch server host
+      # @option config [Object] :submit_host (nil) The batch server host
       # @option config [Object] :exec (nil) Path to PBS Pro executables
       # @option config [Object] :qstat_factor (nil) Deciding factor on how to
       #   call qstat for a user
       # @option config [#to_h] :bin_overrides ({}) Optional overrides to PBS Pro client executables
       def self.build_pbspro(config)
         c = config.to_h.compact.symbolize_keys
-        host = c.fetch(:host, nil)
+        submit_host = c.fetch(:submit_host, "")
```
ericfranz, line 19: We want to keep both host and submit host as they are different values.

- `host` is the host of the batch server we are submitting the job to
- `submit_host` is the host where we are executing the job submission command

ericfranz:
I mentioned one example inline, but for each adapter we want to keep both host and submit host as they are different values.

- `host` is the host of the batch server we are submitting the job to
- `submit_host` is the host where we are executing the job submission command

For example, I could specify the submit_host as being ruby.osc.edu and the host as being owens.osc.edu. Then I would be ssh-ing to ruby.osc.edu and executing qsub with the batch server specified as owens.osc.edu to submit a batch job running on owens.

lib/ood_core/job/adapters/lsf/batch.rb
```diff
@@ -141,6 +141,7 @@ def cluster_args
     # Call a forked Lsf command for a given cluster
     def call(cmd, *args, env: {}, stdin: "")
       cmd = OodCore::Job::Adapters::Helper.bin_path(cmd, bindir, bin_overrides)
+      cmd = OodCore::Job::Adapters::Helper.ssh_wrap(cmd, submit_host)
```
ericfranz, line 144: When calling `o, e, s = Open3.capture3(env, cmd, *args, stdin_data: stdin.to_s)`, when executing via ssh it would be preferable if cmd is be `ssh` and args has the rest of the arguments, one per array element. This may avoid some potential problems with placing quotes in the correct place for arguments.

Instead of wrapping cmd to build a larger string, you would have code that does something similar to what the Linux host adapter does. Though that adapter builds an array of arguments, where the first element is the command, and then it passes it to call. Here, you will want to potentially produce new cmd and args (instead of just cmd). One approach is to use multiple assignment for that:

```
irb(main):001:0> a,b = [1,2]
=> [1, 2]
irb(main):002:0> a
=> 1
irb(main):003:0> b
=> 2
```

lib/ood_core/job/adapters/helper.rb
```diff
@@ -12,6 +12,15 @@ class Helper
...
           bin_overrides.fetch(cmd.to_s) { Pathname.new(bin_default.to_s).join(cmd.to_s).to_s }
         end
+        
+        # Gets a command that submits command on another host via ssh
+        # @param cmd [String] the desired command to execute on another host
+        # @param submit_host [String] where to submit the command
+        # @return [String] command wrapped in ssh if submit_host is present
+        def self.ssh_wrap(cmd, submit_host)
+          return cmd if submit_host.empty?
+          "ssh -t -o BatchMode=yes -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no #{submit_host} \"#{cmd}\""
```
ericfranz, line 22: See comment below about changing this to an array.

Try using -T instead of -t - you will need to test to confirm.

Also, we do not want the default to be `UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no`. If we want to support this option we need to add a strict_host_checking configuration option to the adapters (default to true) and pass that as an argument to this method. See the LinuxHostAdapter implementation of `ssh_cmd`. We might have a solution https://github.com/OSC/ood-documentation/issues/273 where, when the OnDemand web node is properly configured, strict host checking is always used.

matthu017: -t was used for tty allocation for the interactive apps, but in this case neither -t nor -T are needed

lib/ood_core/job/adapters/helper.rb
```diff
@@ -12,6 +12,23 @@ class Helper
...
+        
+        # Gets a command that submits command on another host via ssh
+        # @param submit_host [String] where to submit the command
+        # @param cmd [String] the desired command to execute on another host
+        # @param cmd_args [Array] arguments to the command specified above
+        #
+        # @return cmd [String] command wrapped in ssh if submit_host is present
+        # @return args [Array] command arguments including ssh_flags and original command
+        def self.ssh_wrap(submit_host, cmd, cmd_args)
+          return cmd, cmd_args if submit_host.empty?
```
ericfranz, line 24: nil.empty? crashes. Protect against nil by coercing to string, since nil.to_s == "":

```
return cmd, cmd_args if submit_host.to_s.empty?
```

lib/ood_core/job/adapters/helper.rb
```diff
@@ -12,6 +12,23 @@ class Helper
...
+          
+          args = ['-o', 'BatchMode=yes', '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=yes', "#{submit_host}"]
```
ericfranz, line 26: I believe the default is yes so we could omit `'-o', 'StrictHostKeyChecking=yes'` or we could keep it but regardless we want to make `yes` configurable to `no` based on strict_host_checking=false for the adapters (see linux host adapter for an example)

johrstrom:
With the realization in #201 we may have to rethink how the `job_environment` works for this feature. 

Instead of just passing the command, we may have to template a small script that can inline the correct environment variables (like the linux host adapter does). I don't believe we want to get involved with `SendEnv` and `AcceptEnv` just because that would require admins to go through some pain, then realize there's missing sshd configuration.

Let's look at how this would work for PBSPro.
`job_environment` is OK because it's a part of the command.
https://github.com/OSC/ood_core/blob/9eaf3cfe937657104f7d16b4bf2452be18cbdebc/lib/ood_core/job/adapters/pbspro.rb#L262-L263

But these two environment variables need to be passed to the destination. Anything that is set in the environment to the `call` or `Open3` functions needs to be passed to the `submit_host`.
https://github.com/OSC/ood_core/blob/9eaf3cfe937657104f7d16b4bf2452be18cbdebc/lib/ood_core/job/adapters/pbspro.rb#L164-L165

Maybe instead of templating a script (because that would require all the adapters to toggle the `cmd` or `stdin` - the script), we just inline the variables in the command.

Something to this affect: 
`ssh johrstrom@owens.osc.edu 'export FOO=BAR; echo $FOO'`

The `ssh_wrap` interface probably expands to include environment you want to pass, and you prepend exports to it.

So submitting to owens with PBSPro adapter ends up resulting in a command like this:
`ssh johrstrom@owens.osc.edu 'export PBS_DEFAULT=owens; qsub myscript.sh'`

(Note that the quotes are important becuase the shell will see the `;` as the terminating command which it isn't. At least not on the source side).

johrstrom:
Also, I thought adapters like Torque had some restriction because they use shared libraries (FFI) that may not be on the destination hosts?

ericfranz:
Torque does not use FFI for qsub

johrstrom:
Is that the behavior we want, where some things are executing on the submit_host and others on the webhost?

ericfranz:
https://github.com/OSC/ood_core/issues/170 Addresses that. This PR is the first half, enabling only qsub by specifying `submit_host: "some_other_host"`. Another feature would be adding `use_submit_host_for_all_commands: true` to enable the rest of the commands executed on the submit host. In Torque and SGE's cases this second step is problematic because of their use of FFI. We could drop FFI then.

spec/job/adapters/slurm_spec.rb
```diff
@@ -1106,6 +1106,40 @@ def job_info(opts = {})
...
+      end
+    end
+
+    context "when calling with submit_host & strict_host_checking not specified" do 
+      it "uses ssh wrapper & host checking defaults to true" do 
+        batch = OodCore::Job::Adapters::Slurm::Batch.new(cluster: "owens.osc.edu", conf: "/etc/slurm/conf/", bin: nil, bin_overrides: {}, submit_host: "owens.osc.edu")
+        allow(Open3).to receive(:capture3).and_return(["job.123", "", double("success?" => true)])
+
+        OodCore::Job::Adapters::Slurm.new(slurm: batch).submit script
+        expect(Open3).to have_received(:capture3).with(anything, 'ssh', '-o', 'BatchMode=yes', '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=yes', 'owens.osc.edu', 'sbatch --export NONE --parsable -M owens.osc.edu', any_args)
```
ericfranz, line 1128: Why do we do

    'owens.osc.edu', 'sbatch --export NONE --parsable -M owens.osc.edu'

instead of

    'owens.osc.edu', 'sbatch', '--export', 'NONE', '--parsable', '-M', 'owens.osc.edu'

lib/ood_core/job/adapters/helper.rb
```diff
@@ -12,6 +12,31 @@ class Helper
...
+        def self.ssh_wrap(submit_host, cmd, cmd_args, strict_host_checking = true, env = {})
+          return cmd, cmd_args if submit_host.to_s.empty?
+
+          check_host = strict_host_checking ? "yes" : "no"
+          args = ['-o', 'BatchMode=yes', '-o', 'UserKnownHostsFile=/dev/null', '-o', "StrictHostKeyChecking=#{check_host}", "#{submit_host}"]
+
+          command = ""
+          env.each {|key, value| command += "export #{key}=#{value}; "} # exports env vars
+          command += cmd_args.unshift(cmd).join(' ')
+          args.push(command)
```
ericfranz, line 34: This does the same thing:

```
args.push(env.map {|k,v| "export #{key}=#{value};" } + [cmd] + cmd_args).join(' '))
```

though `env.map {|k,v| "export #{key}=#{value};" }` should be a separate method

spec/job/adapters/slurm_spec.rb
matthu017: In bash, something like `'touch foo.txt; echo foo > foo.txt'` would only work with quotations. However ruby's open3 it works as an array instead as one string. (Plus we aren't using multiple commands)
Will change back to an array for args.

lib/ood_core/job/adapters/helper.rb
matthu017: Should a new method still be made for this with the last change? There are quite a few params in this method already, but everything in this command deals with ssh wrapping.

lib/ood_core/job/adapters/helper.rb
```diff
@@ -12,6 +12,27 @@ class Helper
...
+        #
+        # @return cmd [String] command wrapped in ssh if submit_host is present
+        # @return args [Array] command arguments including ssh_flags and original command
+        def self.ssh_wrap(submit_host, cmd, cmd_args, strict_host_checking = true, env = {})
+          return cmd, cmd_args if submit_host.to_s.empty?
+
+          check_host = strict_host_checking ? "yes" : "no"
+          args = ['-o', 'BatchMode=yes', '-o', 'UserKnownHostsFile=/dev/null', '-o', "StrictHostKeyChecking=#{check_host}", "#{submit_host}"]
+          env.each{|key, value| args.push("export #{key}=#{value};")}
+          args.concat(cmd_args.unshift(cmd))
```
ericfranz, line 31: `cmd_args.unshift(cmd)` modifies the argument `cmd_args` which should not happen. I know you just changed a bunch of + to concats, but in this case + may improve readability.


    return 'ssh', args + [cmd] + cmd_args

PR #205, opened 2020-06-23
Added qos option to torque and slurm

matthu017:


matthu017:
I couldn't find any definitive sources that PBSpro uses qos

ericfranz:
Tests are needed for this. See how tests for accounting_id for example are done.

Issue #149, opened 2019-08-03
squeue timeout reports as job being completed


ericfranz:
See comment: https://github.com/OSC/ood-dashboard/issues/171#issuecomment-412954650

ericfranz:
To reproduce this issue, we would need to find out the exact output that squeue produces after the timeout, and then alter the code to handle this situation, and not confuse it with the job being completed.

johrstrom:
Looking online it appears `slurm_load_jobs error: Socket timed out on send/recv operation` is what's returned out of std error (in Kristina's case in the linked issue) and `CLUSTER: amarel` out of standard out.  `exit_group(0)` means it exists with exit code 0, so that's why no error was thrown up the stack (to avoid this issue).  

We can likely just mock this output in a test case. The solution is probably to distinguish between standard error and std out and throw an error if there's anything in standard error (though isn't debug output in standard error?).  Or to better interpret the output, checking for error strings if they're regular enough (and throwing the error). 

That one output string is given, but we should also search and try to see if there are other timeout/error strings we can look for too.

johrstrom:
This was fixed in #209

PR #209, opened 2020-09-09
slurm handle squeue timeouts

johrstrom:
Possible fix for #149.

I don't really like using exceptions for control flow, but the way it's structured currently I'm not sure how else to do with without a much more invasive change.


johrstrom:
@ericfranz are you good with this?

PR #212, opened 2020-10-01
Lha timing bug

### LINKED ISSUE #211: latency in bashr causes LHA to complete immediatley
We seem to have some race condition in the LHA here where we delete the temporary files too quickly.  To replicate simply add a `sleep 1` in your `~/.bashrc` and a job will complete immediately. Of course nobody is really sleeping in their bashrc, but _am_ loading some things that cause latency.

What's worse is it's not immediately apparent what's going on because we default error_path to /dev/null and batch connect apps only populate output_path so I only found this out by modifying the code itself.

johrstrom:
Fixes #211 

Initially I'd added some complexity to look into the tmux pstree to see if singularity had started, but now I'm just thinking why not just keep the files in the `workdir`?  The `debug` flag won't really do much now, but there's no race condition and it'll mean we can always debug.


ericfranz:
LGTM

PR #216, opened 2020-11-20
Add 'account' label to K8 pods

treydock:


lib/ood_core/job/adapters/kubernetes/templates/pod.yml.erb
```diff
@@ -7,6 +7,7 @@ metadata:
     job: <%= id %>
     app.kubernetes.io/name: <%= container.name %>
     app.kubernetes.io/managed-by: open-ondemand
+    account: <%= script.accounting_id %>
```
johrstrom, line 10: We should probably account for nil here. Otherwise it would be `account: ` which would probably throw a yaml parsing error by kubectl which is particularly unreadable to users.

treydock: I updated to guard against `nil` `account_id`.

treydock:
@johrstrom Added commit to git rid of all the extra newlines in k8 template.  Figured at very least if the template persists for debugging it would make it easier to read.

lib/ood_core/job/adapters/kubernetes/templates/pod.yml.erb
johrstrom: Cool. I pulled the other PRs in (Thank you for them!).  I just need to test this to be sure that `account: -` works.

treydock: Why would there be a hyphen for `account`?  Also for at least nodes empty labels are valid so I would imagine empty labels are fine. In Go an undefined string is an empty string and if their labels are `map[string]string` then no value is same as `key: ""`.

treydock: I think any valid string for the value will be fine assuming the Go code for labels defines it as `map[string]string`, I think the only way it would not work is if they define labels as `map[string]interface` but I doubt they do that.

johrstrom: Am I misreading `ERB.new(File.read(resource_file), nil, '-')`? I read that as replace `nil` values with `-`. 

In any case, the comment was just to update you that I want to put it through an end to end test just to be sure.

treydock: That `'-'` in `ERB.new` instructs ERB rendering to strip newlines when `-%>` is used and strip leading spaces with `<%-` is used.  See the change to test rendered YAML, all the extra newlines are no longer present. So that something like this:

```
   <%- if foo -%>
bar
   <%- end -%>
```

WIll render just `bar` and not:

```
   
bar

```

treydock:
@johrstrom Does this look okay to merge?  Soon I will deploying the policy engine at OSC that will prevent OnDemand pods from starting without `account` label but also requires the namespace prefix.

johrstrom:
Yea I haven't got a chance to try end to end with OOD yet. But I tried a simple container with nothing in account field and it errored.  I can share the yml in case you don't have one handy to test the same with.

```yaml
[testing()] 🐮  head -n 10 bash_cpu.yml 
apiVersion: v1
kind: Pod
metadata:
  namespace: johrstrom
  name: bash
  labels:
    foo: 'bar'
    account:
spec:
  restartPolicy: OnFailure
```

and got

```
[testing()] 🐢 kubectl create -f bash_cpu.yml 
error: error validating "bash_cpu.yml": error validating data: unknown object type "nil" in Pod.metadata.labels.account; if you choose to ignore these errors, turn validation off with --validate=false
```

treydock:
But this change won't leave empty `account` field. If `account` is nil it will be omitted from template.  I can further add safety measure to leave off empty strings too but a nil account field isn't possible with `if` statement used.

treydock:
@johrstrom Updated template logic so `nil` or empty string will prevent account label from getting set.

johrstrom:
Cool, I hadn't checked it out since that if block was added. That's fine. Honestly just quoting it would have been fine with me like `account: '<%= script.accounting_id %>'` but we can pick that up in #223.

I'll pull this in once the tests run (if they complete in a reasonable amount of time).

PR #221, opened 2020-12-08
replace travis ci with gh actions

johrstrom:
replace travis ci with gh actions. This is just testing now, but we should look into publishing the gem automatically too afterwards.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,32 @@
+name: Unit Tests
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - uses: actions/setup-ruby@v1
+        with:
+          ruby-version: '2.5'
+      - uses: actions/cache@v2
+        with:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          ruby-version: '2.5'
+      - uses: actions/cache@v2
+        with:
+          path: ~/vendor/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('Gemfile.lock') }}
+      - name: setup bundler
+        run: |
+          gem install bundler -v 1.17.3
+          bundle config path ~/vendor/bundle
```
treydock, line 28: Why all the extra bundler logic?  I don't know if this would solve the problem but setting up bundle gem location and all that shouldn't be necessary.  This is example what I'm doing in other repos: https://github.com/treydock/puppet-module-keycloak/blob/master/.github/workflows/ci.yaml#L45-L49.  The `setup-ruby` action will do the bundle install

johrstrom: It's so the dependency gems can be cached.  There are not a lot with ood_core, but I took this from the OOD PR for the same which does have a lot of gems we should be caching.

https://github.com/OSC/ood_core/blob/70d7ff537483592de1f7b19e6c7e2e7591aff238/.github/workflows/test.yml#L21-L24

PR #222, opened 2020-12-08
specify k8s connection info in attr_reader of new K8sJobInfo

johrstrom:
I believe we're going to have to utilize connection information in native for kuberenetes.  So this PR is to specify that hash within native so that it doesn't conflict with what other schedulers may populate as `native.host` or `native.port`.

Note this commit also adds the RAILS-LICENSE because the addition of deep_merge(!) in HashExtentions. But given that we already seem to use their slice and compact it seemed appropriate to add their
license.


johrstrom:
I think the logic of having `native.ood_connection_info` in and of itself is OK. The only question I'd have is about adding `deep_merge` and the Rails license.

ericfranz:
We could just have ood_core have a very loose dependency on active_support - like active_support with a range of versions between the active_support that is used in Rails 5 and 6 (maybe just active_support >= whatever version we know about). There are not too many places we use ood_core that is not a Rails app.

johrstrom:
In gemspec is that `add_development_dependency`?

johrstrom:
We use it for reporting queue times. Also it seems like we should be using it in https://github.com/OSC/osc-systemstatus - a Sinatra app.

I mean I can add it as a `development_dependency` to the gemspec, but the k8s adapter just wouldn't work outside of Rails which I feel like is kind of unexpected.

ericfranz:
No it would be a runtime dependency. You can use active_support outside of Rails. It is a standalone gem.

johrstrom:
That works for me.  I used `spec.add_runtime_dependency "activesupport", ">= 5.2", "< 6.0"` just so we'll just need to remember to update this when we upgrade to rails 6.

ericfranz:
Do we only need host right now, or will we need more information in the future? If we only need host, another option would be to just utilize "allocated_nodes". The name of the node could be this "host" string. I guess the procs part of NodeInfo would be irrelevant, and even though "allocated_nodes" is an array, for K8s you could assume you are connecting to the first host.

If we really need a full connection object or a bunch of connection attributes, here is another option. Since in this PR we are overloading the "native" attribute for this information, we are doing the equivalent of:

```
class K8sJobInfo < OodCore::Job::Info
  attr_accessor :connection_info
end

# ... then in the dashboard:
def connect
  if info.respond_to?(:connection_info)
    OpenStruct.new(info.connection_info || {})
  else
    OpenStruct.new YAML.safe_load(connect_file.read)
  end
end
```

However, a Job Info subclass or a new attribute on OodCore::Job::Info would avoid using native hash for this case and might be a little more explicit. Since this is the first time that we have had a need for special use of the native attribute to actually change the behavior (or do we do that in the linux host adapter at all?) maybe a subclass of OodCore::Job::Info is preferable to start with.

johrstrom:
Actually, `Kubernetes::K8sJobInfo` was probably the right direction.

johrstrom:
Rebased. I think I'm ready for this if you have anything else?

PR #224, opened 2020-12-08
Support defining pod lifetime for Kubernetes

treydock:
I opted to use easier to read format like `1h5m` rather than just do everything as seconds.  If preferred could just as easily make everything seconds, so 3600 would become `3600s` (`s` suffix).

Also verified `annotations` with no key/value pairs does work.

### DIFF
--- a/lib/ood_core/job/adapters/kubernetes/templates/pod.yml.erb
+++ b/lib/ood_core/job/adapters/kubernetes/templates/pod.yml.erb
@@ -10,6 +10,10 @@ metadata:
     <%- if !script.accounting_id.nil? && script.accounting_id != "" -%>
     account: <%= script.accounting_id %>
     <%- end -%>
+  annotations:
+    <%- unless script.wall_time.nil? -%>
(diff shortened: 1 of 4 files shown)


lib/ood_core/job/adapters/kubernetes/templates/pod.yml.erb
```diff
@@ -10,6 +10,10 @@ metadata:
     <%- if !script.accounting_id.nil? && script.accounting_id != "" -%>
     account: <%= script.accounting_id %>
     <%- end -%>
+  annotations:
+    <%- unless script.wall_time.nil? -%>
+    pod.kubernetes.io/lifetime: <%= seconds_to_duration(script.wall_time) %>
```
johrstrom, line 15: This needs to be `helper.seconds_to_duration(script.wall_time)` otherwise we get undefined method error (because the templating context is the batch class).

treydock: Made that change

PR #230, opened 2020-12-11
Unscheduleable is queued

### LINKED ISSUE #229: k8s could handle Pending/queued state better?
Right now, in ood_core for k8s if you're in the k8s `Pending` state you can map to undetermined or queued.  It doesn't seem like it should map to undetermined. 

Right now the 2 Pending maps to
* `undetermined` when you've not started at all (let's say you've requested some resources that aren't available yet).
  * initially I think I made this `undetermined` because k8s will accept requests for the impossible and you're forever in this Pending state.
* `queued` when you are starting (you're scheduled on a node and so on) and likely init containers are running.

I'm guessing both should map to queued? and the dashboard has to figure out a better way to determine `starting` state.

https://github.com/OSC/ood_core/blob/f89a3005379618b7f04743fe2b573ebfdf50a664/lib/ood_core/job/adapters/kubernetes/helper.rb#L256-L273

johrstrom:
Fixes #229 

Unscheduleable pods will now get a status of `queued`, which they are kind of.  You would either be unscheduleable (wrong node selector or impossible amount of resources) or you're actually waiting for some resources to free up with is valid.


johrstrom:
I think putting `Pending` -> `queued` is the right move. You're waiting on resources to free up or the init containers are running.

If they never become free, well then you'll have to investigate, but at least it won't cause false positives like `undetermined` would if you legitimately wait for a few minutes.

ericfranz:
I guess there is no way to determine if a pod is unscheduleable if it is stuck in the pending state?

ericfranz:
Maybe we go with this solution but then post an issue to find a solution to determine pods stuck in an unschedulable state. One example of a solution we could explore could be a simple one - if a pod has been stuck in an unschedulable state for 24 hours (or some arbitrary time span) it changes to "undetermined" for example.

johrstrom:
That works for me.

treydock:
There are ways to see pod is not able to run.  Also with the `job-pod-reaper` if a pod is created with lifetime of say `8h` and never starts, it will get killed after 8 hours.

PR #231, opened 2020-12-11
Helper namespace

### LINKED ISSUE #219: clean up namespaces in k8s adapter
In #218 we added a prefix to namespaces and it reminded me that we could use a small refactor here.

We probably don't need `default_namespace` here. There's no real reason for it here that I can discern, it doesn't really add any value.
https://github.com/OSC/ood_core/blob/6eb564a919c53b36559bcbbb27c46af6e409cb92/lib/ood_core/job/adapters/kubernetes/batch.rb#L207-L213

So this ticket is clean that up a bit and maybe add a test case for setting the namespace as well.

johrstrom:
With #218 we now need to convert map the namespace back to `info.job_owner` because their not necessarily the same any more.

This also fixes #219 though that's kinda secondary.

### DIFF
--- a/spec/fixtures/output/k8s/pod_yml_from_defaults.yml
+++ b/spec/fixtures/output/k8s/pod_yml_from_defaults.yml
@@ -8,6 +8,7 @@ metadata:
     app.kubernetes.io/name: rspec-test
     app.kubernetes.io/managed-by: open-ondemand
     account: test
+  annotations:
 spec:
   restartPolicy: Always
   securityContext:
@@ -67,6 +68,8 @@ kind: Service
 metadata:
   name: rspec-test-123-service
   namespace: testuser
+  labels:
+    job: rspec-test-123
 spec:
   selector:
     job: rspec-test-123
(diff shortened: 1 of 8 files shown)


spec/fixtures/output/k8s/pod_yml_from_defaults.yml
```diff
@@ -0,0 +1,88 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  namespace: testuser
+  name: rspec-test-123
+  labels:
+    job: rspec-test-123
+    app.kubernetes.io/name: rspec-test
+    app.kubernetes.io/managed-by: open-ondemand
+    account: test
```
treydock, line 10: I think with #224 there will be empty `annotations:` so tests may fail without that extra metadata key being here.

johrstrom: Thanks! Yea I needed a rebase because merge conflict too, I think because that PR.

Anyhow - yea, I rebased and had to merge that.

Issue #182, opened 2020-03-09
User level configuration


johrstrom:
I've been seeing tickets all over the place that would require user level configurations, and I think it'd be a really slick feature. 

I think json files in XDG directories would be a good place to start.

`~/.config/ondemand/apps/some_app.json`

and it's contents are valid json and app specific.  The API may be just like json, `dig(:foo, :bar)` with the app needing to set defaults statically.

ericfranz:
Is this proposal for new user configuration options that we do not have a solution for yet? Maybe YAML is more appropriate since that is what we use elsewhere, though presumably we won't be encouraging users to hand modify this config file?

johrstrom:
Yea so I think this solves configurations specific to the user, while also avoiding an extra UI field or extra DB entry. So it's extendable and avoids cluttering the UI of a given app.

Something like [linking to someone else's templates](https://github.com/OSC/ondemand/issues/165) or [customizable directory paths](https://github.com/OSC/ondemand/issues/176) just to name two. 

```yaml
active_jobs:
    extra_template_dirs: "/foo/bar,/home/professor"
    job_dir: "/home/jeff/scripts"
```

I don't think modifying a file would be that bad or hard for users to get around but longer term maybe provide a simple app (or dashboard pane) to navigate through options, like what VS code does below. But yea, if they want to change their configs, they would have to modify the file - at least initially.

![image](https://user-images.githubusercontent.com/4874123/76342097-5fea4980-62d4-11ea-9a48-732dd012d85f.png)


I'm thinking a class that when given a hash of defaults, returns a hash of user defined configs.

cjkeist:
I vote for some form o user config options!

johrstrom:
Thanks! I think the better way is to add a thumbs up reaction at the top.

johrstrom:
We're closing this now as this work has started in OnDemand itself instead of in this library. You can see the module here that handles this work and indeed has already started to store a few things.  This will likely be expanded over time to allow for further user level customizations.


https://github.com/OSC/ondemand/blob/cc6d25cb5d1d5418db22f71de51622a8c6e21a31/apps/dashboard/app/models/concerns/user_setting_store.rb

PR #236, opened 2020-12-31
User preferences

johrstrom:
Possible fix for #182

Initially I thought about doing all this stuff statically, so then users can re-load when they want. This could cut down on file reads.  I left the commit here https://github.com/OSC/ood_core/commit/b82a9c0320bda237bfc3c7a92072426566d9363f to view.

But then I thought it's just simpler to read the file when you create the object. That way the upper layers can cache the object and create a new one when they need to reload the configs.

### DIFF
--- a/lib/ood_core/user_preferences.rb
+++ b/lib/ood_core/user_preferences.rb
@@ -2,33 +2,29 @@ module OodCore
   class UserPreferences
     require 'yaml'
 
-    class << self
-
-      def preferences(app: nil, reload: false)
-        app.nil? ? _prefs(reload: reload) : _prefs(reload: reload).fetch(app, {})
-      end
-
-      def preferences_file
+    def preferences_file
+      @preferences_file ||= begin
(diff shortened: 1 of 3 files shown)


ericfranz:
Lets discuss offline in more detail. I think we should postpone modification to ood_core with support for user preferences until we complete the launch UI changes that we want to extend user configuration to and merge the core apps that we are able to easily merged. This is for two reasons known up front though there may be others.

1. A simple key/value interface used in the Rails application may be the preferred interface, instead of a hash representation of a hierarchical YAML. For example, we could follow the convention that is used for i18n, where key's use the `.` the separator. Rails offers multiple abstractions and solutions we may be able to leverage as well for this.
2. If we are successful in merging the core apps into one, we will likely be left with a nodejs app (shell) and a dashboard app (Rails) - the modification in ood_core will not provide a benefit of sharing between multiple apps since we can't use ood_core from nodejs and splitting code that works with user settings between ood_core and the dashboard repos before we understand the proper abstraction required will add extra work.

PR #237, opened 2020-12-31
update k8s to safely get and delete resources that no longer exist

johrstrom:
update k8s to safely get and delete resources that no longer exist.  Before things that didn't exist were marked by the dashboard as 'Undetermined' (probably because the errors propagated up to it). And what's worse is errors were thrown trying to _delete_ it too, so you couldn't delete the pod & other resources.

### DIFF
--- a/lib/ood_core/job/adapters/kubernetes/batch.rb
+++ b/lib/ood_core/job/adapters/kubernetes/batch.rb
@@ -11,6 +11,7 @@ class OodCore::Job::Adapters::Kubernetes::Batch
   using OodCore::Refinements::HashExtensions
 
   class Error < StandardError; end
+  class NotFoundError < StandardError; end
 
   attr_reader :config_file, :bin, :cluster, :mounts
   attr_reader :all_namespaces, :using_context, :helper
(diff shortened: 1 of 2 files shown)


treydock:
What will happen if job-pod-reaper deletes things?  Will the delete operation in the code here handle that or produce errors?

johrstrom:
Yea I think that's what prompted this.  There's a bug now, where we _don't_ really handle things that are deleted so this is a bugfix for that.  In fact right now, you can't delete something that is already deleted.

So this addition is to catch when something isn't found.
https://github.com/OSC/ood_core/blob/e89119a1bf2c71ed4f2a40526036090a5e663109/lib/ood_core/job/adapters/kubernetes/batch.rb#L352-L355

And rescue that, so if it can't find what it's looking for to delete _or_ get, there's nothing to do and we can just gracefully return nothing.
https://github.com/OSC/ood_core/blob/e89119a1bf2c71ed4f2a40526036090a5e663109/lib/ood_core/job/adapters/kubernetes/batch.rb#L123-L134

PR #250, opened 2021-03-16
Kubernetes adapter updates

treydock:
* Rework how configmap is defined for several purposes
  * Support changing mountPath for different files in configmap
  * Support mounting subPath files in configmap
  * Support mounting files that might not be defined statically but instead generated in init container
* Add some helper environment variable to both main and init container

### DIFF
--- a/lib/ood_core/job/adapters/kubernetes/resources.rb
+++ b/lib/ood_core/job/adapters/kubernetes/resources.rb
@@ -1,12 +1,28 @@
 module OodCore::Job::Adapters::Kubernetes::Resources
+  require_relative 'batch'
 
   class ConfigMap
-    attr_accessor :name, :filename, :data
+    attr_accessor :name, :files
 
-    def initialize(name, filename, data)
+    def initialize(name, files)
       @name = name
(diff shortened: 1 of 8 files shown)


lib/ood_core/job/adapters/kubernetes/resources.rb
```diff
@@ -1,12 +1,28 @@
 module OodCore::Job::Adapters::Kubernetes::Resources
+  require_relative 'batch'
```
johrstrom, line 2: I don't think we need this require here, seems like it'd be cyclical.

johrstrom:
Can we break this up into 2 PRs?  One for the configmap updates and the other for the environment updates.

I think the configmap updates look OK, though I just need to lookup the k8s documentation on those fields for my own edification.  But existing test cases didn't need a lot of updates, so that's a good sign.

treydock:
Will need to merge configmap updates first and then the env updates will have to get updated to ensure the new test file has appropriate env values for testing.  I opened #251 for confgmap changes, will open another PR for environment variable changes.

PR #252, opened 2021-03-17
Add some helper environment variable to both main and init container

treydock:
Replaces new env variables from #250 with suggested changes so that sites can override the defaults.  Also I set it so the default env for init container is the env of main container which figured might make overriding defaults a bit more DRY.

### DIFF
--- a/lib/ood_core/job/adapters/kubernetes/helper.rb
+++ b/lib/ood_core/job/adapters/kubernetes/helper.rb
@@ -38,18 +38,21 @@ class OodCore::Job::Adapters::Kubernetes::Helper
   #
   # @param container [#to_h]
   #   the input container hash
+  # @param default_env [#to_h]
+  #   Default env to merge with defined env
   # @return  [OodCore::Job::Adapters::Kubernetes::Resources::Container]
-  def container_from_native(container)
+  def container_from_native(container, default_env)
     OodCore::Job::Adapters::Kubernetes::Resources::Container.new(
       container[:name],
       container[:image],
       command: parse_command(container[:command]),
       port: container[:port],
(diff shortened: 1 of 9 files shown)


treydock:
Also env is now a key/value pair rather than array of hashes.

treydock:
Either #251 gets merged before this and then this PR updated or other way around, otherwise the tests added in #251 will begin to fail if not updated with env changes.

lib/ood_core/job/adapters/kubernetes/helper.rb
```diff
@@ -38,18 +38,21 @@ def info_from_json(pod_json: nil, service_json: nil, secret_json: nil, ns_prefix
...
       command: parse_command(container[:command]),
       port: container[:port],
-      env: container.fetch(:env, []),
+      env: container.fetch(:env, {}),
       memory: container[:memory],
       cpu: container[:cpu],
       working_dir: container[:working_dir],
-      restart_policy: container[:restart_policy]
+      restart_policy: container[:restart_policy],
+      default_env: default_env
```
johrstrom, line 55: `deep_merge` is already available to the helper. Seems like we shouldn't have a `default_env` member variable and instead just merge when creating `@env`. That way we can get rid of the `env` method on the resource too.

https://github.com/OSC/ood_core/blob/087d28596c85a6ec77000a1e2045ba6e218acb63/lib/ood_core/job/adapters/kubernetes/helper.rb#L7

treydock: I updated to use `merge` since these are not going to be deep hashes, only key/value pairs.

johrstrom (approved):
LGTM

PR #255, opened 2021-03-25
Make K8 configmap configuration more intuitive

### LINKED ISSUE #254: make k8s configmap configuration a bit more intuative
With the addition of #251 hard coding the configmap mount path to `/ood` seems a bit clunky and makes configuration a bit counter intuitive in 

So this ticket is to get rid of this method and instead force the specification of the configmap mount path https://github.com/OSC/ood_core/blob/c6c833d343931521428d2f2bfddb8ed2c3468843/lib/ood_core/job/adapters/kubernetes/batch.rb#L115-L117. 

So a config like this would a configmap that would generate a file at the given directory for both init containers and the real container.
```yaml
configmap:
  files:
    - filename: "myfilename"
      mount_path: '/ood/myfile'
      data: |
        my file's data
```

This configuration would specify a different place to mount in the init containers.
```yaml
configmap:
  files:
    - filename: "myfilename"
      mount_path: '/ood/myfile'
      init_mount_path: '/init/my_init_file'  # specify a different init path
      data: |
        my file's data
```

I also wonder if we should allow for `init_mount_path: false`.

https://github.com/OSC/ood_core/blob/c6c833d343931521428d2f2bfddb8ed2c3468843/lib/ood_core/job/adapters/kubernetes/templates/pod.yml.erb#L109

It seems more intuitive to just instruct folks to use an init mount location that they'll ignore like `init_mount_path: '/wont_be_used'`.  Sort of like redirecting to /dev/null.  Instead of yet another configuration variant, simply ignore the init mount path.

treydock:
Fixes #254 

I opted so that the init configmap mount only happens when you define `init_mount_path`, same for main container and `mount_path`.

### DIFF
--- a/spec/fixtures/output/k8s/pod_yml_no_mounts.yml
+++ b/spec/fixtures/output/k8s/pod_yml_no_mounts.yml
@@ -47,9 +47,6 @@ spec:
     - "spec"
     ports:
     - containerPort: 8080
-    volumeMounts:
-    - name: configmap-volume
-      mountPath: /ood
     resources:
       limits:
         memory: "6Gi"
@@ -87,9 +84,6 @@ spec:
     - "/bin/ls"
     - "-lrt"
     - "."
-    volumeMounts:
-    - name: configmap-volume
-      mountPath: /ood
     securityContext:
       allowPrivilegeEscalation: false
       capabilities:
--- a/lib/ood_core/job/adapters/kubernetes/batch.rb
+++ b/lib/ood_core/job/adapters/kubernetes/batch.rb
@@ -112,10 +112,6 @@ class OodCore::Job::Adapters::Kubernetes::Batch
     safe_call("delete", "configmap", configmap_name(id))
   end
 
-  def configmap_mount_path
-    '/ood'
-  end
-
   private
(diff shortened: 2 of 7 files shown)


spec/fixtures/output/k8s/pod_yml_no_mounts.yml
```diff
@@ -47,9 +47,6 @@ spec:
     - "spec"
     ports:
     - containerPort: 8080
-    volumeMounts:
-    - name: configmap-volume
-      mountPath: /ood
```
johrstrom, line 52: This test file/test case was added #239 as a bugfix. You can see in the yaml it still has the volume, though we get rid of the volume mount.

Maybe we need an additional test case where this handles "no volumes or mounts whatsoever" and another that handles "no volumes but we do mount a configmap".

treydock: I made this particular test case go back to previous behavior and added new test case with no mounts and no configmap mounts.

johrstrom (approved):
Thanks! I think I'm good with this.

PR #258, opened 2021-03-29
tmux 2.7 does not like non uni-code field seperators

johrstrom:
A bug report from discourse:

https://discourse.osc.edu/t/ood-as-a-teaching-tool/1401

Turns out tmux 2.7 (and possibly earlier, but 2.7 ships with centos:8) does not like `"\x1F"` as a separator and instead prints underscores. So when we go to parse it, we can't parse it correctly because it cannot split on `\x1F`.

So this simply changes the field separator to commas which are not special characters in bash, so I think they're OK to use.


ericfranz:
I don't think you intended `spec/fixtures/output/k8s/single_queued_w_running_init_ctrs_pod.json` to be included in the PR

PR #259, opened 2021-04-02
interpret tmux 1.8 and 2.7 error messages

johrstrom:
tmux error messages changed at some point in 2.x. This was confirmed in 2.7 (what ships with centos/rhel 8). This also removes a test case and error message that's no longer valid. We don't use the tmux 'has-session' api anywhere and therefore should not check for it's error messages.

While #258 fixes an issue found with 2.7, this updates so that we can interpret 2.7's error messages as well.

### DIFF
--- a/lib/ood_core/job/adapters/linux_host/launcher.rb
+++ b/lib/ood_core/job/adapters/linux_host/launcher.rb
@@ -80,12 +80,7 @@ class OodCore::Job::Adapters::LinuxHost::Launcher
 
     call(*cmd, stdin: kill_cmd)
   rescue Error => e
-    raise e unless (
-      # The tmux server not running is not an error
-      e.message.include?('failed to connect to server') ||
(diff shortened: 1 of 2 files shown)


lib/ood_core/job/adapters/linux_host/launcher.rb
```diff
@@ -80,12 +80,7 @@ def stop_remote_session(session_name, hostname)
 
     call(*cmd, stdin: kill_cmd)
   rescue Error => e
-    raise e unless (
-      # The tmux server not running is not an error
-      e.message.include?('failed to connect to server') ||
-      # The session not being found is not an error
-      e.message.include?("session not found: #{session_name_label}")
```
johrstrom, line 87: This error message, and the corresponding test case are unused. They come from running `tmux has-session -t the-session-im-looking-for`, but we don't use that API anywhere.

PR #262, opened 2021-04-16
Allow K8 pods to select a specific compute cluster

treydock:
The K8 environment at OSC will have some worker nodes with `cluster=pitzer` from that compute environment and some with `cluster=owens`.  It's highly likely they will have differences if nothing else than the hardware.  If we don't care about the hardware a user can request then this pull request serves no purpose.  For things like /apps and /usr/local I think what we'd do is mount pitzers /apps to /apps/pitzer and owens to /apps/owens, we can't mount things to /usr/local.

### DIFF
--- a/lib/ood_core/job/adapters/kubernetes/templates/pod.yml.erb
+++ b/lib/ood_core/job/adapters/kubernetes/templates/pod.yml.erb
@@ -148,6 +148,10 @@ spec:
   <%- end # if mount is [host,nfs] -%>
   <%- end # for each mount -%>
   <%- end # (configmap.to_s.empty? || all_mounts.empty?) -%>
+  <%- unless cluster.nil? -%>
+  nodeSelector:
+    cluster: <%= cluster %>
+  <%- end -%>
 ---
 <%- unless spec.container.port.nil? -%>
 apiVersion: v1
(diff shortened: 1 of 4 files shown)


lib/ood_core/job/adapters/kubernetes/templates/pod.yml.erb
```diff
@@ -148,6 +148,10 @@ spec:
   <%- end # if mount is [host,nfs] -%>
   <%- end # for each mount -%>
   <%- end # (configmap.to_s.empty? || all_mounts.empty?) -%>
+  <%- unless cluster.nil? -%>
+  nodeSelector:
+    cluster: <%= cluster %>
+  <%- end -%>
```
johrstrom, line 154: Instead of implementing cluster specifically, seems like we should allow for `nodeSelector` configs. Then we at OSC can use `cluster`, but other folks will be able to use whatever they like.

treydock: Ya, that would actually work out to halfway solve the problem of supporting GPUs since a GPU request is in resources but you can choose which kind of GPU you get via nodeSelector.  I'll close this and open new pull request to do node selector configuration

PR #266, opened 2021-04-20
Add gpus_per_node support for SLURM, Torque and Kubernetes

### LINKED ISSUE #263: Support GPU requests
It would be useful to have ability to directly request GPUs via the various scheduler adapters.

SLURM:
```
--gpus-per-node=N
```

Torque:
```
-l nodes=X:ppn=Y:gpus=N
```

Kubernetes: https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/
```yaml
resources:
  limits:
    nvidia.com/gpu: 1
  requests:
    nvidia.com/gpu: 1
```

treydock:
Fixes #263


johrstrom:
I can't think of a reason why not? At some point like `qos` we just started a scheme where some adapters accept things from script that others don't.

Without an entirely different interface, I'm not sure we can have a 1 size fits all. So maybe this is as good as it gets with this interface - some adapters accept extra fields like `gpus_per_node` and some don't?

treydock:
I think maybe one improvement would be to update the adapters that don't support it to raise errors if it's used, just to avoid sites maybe trying to use this feature on unsupported adapters.

johrstrom (approved):
:+1:

PR #267, opened 2021-04-23
memoize Cluster#allow? #login_allow? & job_allow?

ericfranz:


johrstrom (approved):
LGTM. I know we talked about dropping ACLs but I wonder if we should add more tests just in case we never do.

PR #271, opened 2021-05-12
Make K8 username prefix behave like namespace prefix

treydock:


johrstrom:
It's a bit worrisome that this doesn't break any tests. We need to add some. Happy to defer that if you need though.

treydock:
I tried, the place where this is called , if you test `call(cmd)` you will end up getting that same function tripping up tests from things called during class initialization I think, so either the function call locations need to change or need to accept that a test on `set_context` is also going to be testing other things that got initialized.

This is what I came up with:

```ruby
  describe '#set_context' do
    let(:batch) do
      batch = described_class.new({})
    end
    before(:each) do
      allow_any_instance_of(OodCore::Job::Adapters::Kubernetes::Batch).to receive(:make_kubectl_config).and_return(nil)
      allow(batch).to receive(:username).and_return('test')
      allow(batch).to receive(:home_dir).and_return('/home/test')
    end
    it 'generates correct command' do
      expected_context_cmd = '/usr/bin/kubectl --kubeconfig=/home/test/.kube/config config set-context --cluster=open-ondemand --namespace=test --user=test'
      batch.send(:set_context)
      expect(batch).to receive(:call).with(expected_context_cmd)
    end
  end
```

```
  1) OodCore::Job::Adapters::Kubernetes::Batch#set_context generates correct command
     Failure/Error: o, e, s = Open3.capture3(env, cmd, stdin_data: stdin.to_s)
     
       Open3 received :capture3 with unexpected arguments
         expected: ({}, "/usr/bin/kubectl --kubeconfig=/Users/tdockendorf/.kube/config config set-cluster open-ondemand --server=https://localhost:8080", {:stdin_data=>""})
              got: ({}, "/usr/bin/kubectl --kubeconfig=/Users/tdockendorf/.kube/config config set-context open-ondemand --cluster=open-ondemand --namespace=test --user=test", {:stdin_data=>""})
       Diff:
       @@ -1,4 +1,4 @@
        [{},
       - "/usr/bin/kubectl --kubeconfig=/Users/tdockendorf/.kube/config config set-cluster open-ondemand --server=https://localhost:8080",
       + "/usr/bin/kubectl --kubeconfig=/Users/tdockendorf/.kube/config config set-context open-ondemand --cluster=open-ondemand --namespace=test --user=test",
         {:stdin_data=>""}]
```

I could not get tests to pass despite looking correct, somewhere something else was calling `set_cluster` and I couldn't get that to stop.

treydock:
Finally got unit tests to work.  There are a lot of things spilling over between tests and could not mock `home_dir` so had to just rely on environment variable which seems all kinds of wrong.

PR #273, opened 2021-05-16
K8 - make batch connect script accessible to container, also make workdir accessible

treydock:


johrstrom:
Would we need this if we got `staged_root` available in the submit.yml?

treydock:
The part for setting up `script.sh`, yes that would still be needed. If we had `staged_root` available to `submit.yml.erb` then the `WORKDIR` environment variable part would not be needed though might be still useful to set as a helper.  The huge benefit to pulling `script.sh` from the script content is being able to make it accessible to pod so sites have the option of more easily running the job script generated by OnDemand.

lib/ood_core/job/adapters/kubernetes/templates/pod.yml.erb
```diff
@@ -52,6 +52,10 @@ spec:
       valueFrom:
         fieldRef:
           fieldPath: metadata.namespace
+    <%- unless script.workdir.nil? -%>
+    - name: WORKDIR
+      value: "<%= script.workdir %>"
+    <%- end -%>
```
johrstrom, line 58: Why is `WORKDIR` needed, isn't this just `$PWD`?

treydock: This is only way I could find to get OnDemand staged_root into the pod.  It's not technically needed since I think OnDemand job script will do like `cd <staged_root>` but if that's not always the case then I figured a command like this would be used:

```
cd $(WORKDIR) ; /ood/script.sh
```

It's also the case this is not needed if the script templates all do the `cd <%= staged_root %>`.  Unfortunately to make `PWD` be staged root, would require updating the code in pod.yml.erb to also allow setting working directory to path defined in `script.workdir`.  If Ondemand will always do `cd staged_root` then this environment variable isn't needed.

treydock: I removed `WORKDIR` env variable from this pull request. If we need it in the future it's easy enough to add back. This pull request now just focuses on making job script accessible to pod.

johrstrom:
I think we should fix `staged_root` in https://github.com/OSC/ondemand/issues/864 then make this a little cleaner. That way whatever we merge and deploy will be a little cleaner and easier to use for folks.

treydock:
I took https://github.com/OSC/ondemand/pull/948 and used that to get access to `staged_root` and looks like this:

```
command: ["/bin/bash","-l","<%= staged_root %>/job_script_content.sh"]
```

I think only improvement would somehow making `job_script_content.sh` not something that needs to be hardcoded unless that seems acceptable?

johrstrom:
Yea I think having the script.sh written to a configmap is good. With maybe some sort of opt in or opt out for it? 

It just seems like the `WORKDIR` addition isn't really required if we have staged root. (i.e., we can just set the working directory of the container to `<%= staged_root %>`).  I'm just leery of going one way for immediacies sake then pulling back the very next week.

treydock:
I agree, using staged root for workdir as opt-in makes most sense. I already removed `WORKDIR` addition from this pull request so this pull request is strictly about adding the script to configmap.  How would someone opt-in/opt-out of having script written? In reality is there benefit to opt-out? What's the harm in writing it other than maybe making dump of configmap harder to read.

johrstrom:
Yea i guess to opt-out would be to add a fake `script.sh`?  I guess I'm just thinking about folks who won't have a real `template/script.sh`.  They'll use an actual entrypoint in the container. 

OK, I guess I'm good with this. If it comes up and someone has trouble with it and wants to disable it we'll cross that bridge when we get there.

treydock:
Pushed commit to fix unit tests with recent PRs merged to master branch

johrstrom (approved):
Thanks. I think this'll be a good addition for folks who don't have NFS mounts but still want to be able to use ERB to template stuff and don't want to save lots of things (or need to update stuff) in the container itself.

PR #277, opened 2021-05-17
Save pod yaml into workdir

### LINKED ISSUE #276: Save K8 YAML to OnDemand staged_root for easier debugging
Right now the only way to inspect the YAML for issues is something like `kubectl get pod -n <namespace> <pod name> -o yaml` and that includes a lot of extra information generated by Kubernetes. It would be ideal to save the Kubernetes YAML just like how the current job scripts are saved to staged root.

treydock:
Fixes #276

Screen Shot 2021-05-17 at 8 22 21 AM

### DIFF
--- a/spec/job/adapters/kubernetes/batch_spec.rb
+++ b/spec/job/adapters/kubernetes/batch_spec.rb
@@ -571,66 +571,67 @@ describe OodCore::Job::Adapters::Kubernetes::Batch do
     end
 
     it "saves pod YAML" do
-      tmp = Dir.mktmpdir
-      pod_yml_path = File.join(tmp, "pod.yml")
-      script = build_script(
-        accounting_id: 'test',
-        workdir: tmp,
-        native: {
-          container: {
-            name: 'rspec-test',
-            image: 'ruby:2.5',
-            command: 'rake spec',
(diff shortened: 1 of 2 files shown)


spec/job/adapters/kubernetes/batch_spec.rb
```diff
@@ -568,6 +569,69 @@ def build_script(opts = {})
 
       @basic_batch.submit(script)
     end
+
+    it "saves pod YAML" do
+      tmp = Dir.mktmpdir
+      pod_yml_path = File.join(tmp, "pod.yml")
```
johrstrom, line 575: I think we should pass a block here so we ensure it get's cleaned up.

```ruby
Dir.mktmpdir do |tmp|
  pod_yml_path = File.join(tmp, "pod.yml")
  ...
end
```

johrstrom:
Thanks! I think there's just 1 minor thing with the test case and I'm good with this.

spec/job/adapters/kubernetes/batch_spec.rb
treydock: This change is made, wrapped the entire test block for `Dir.mktmpdir`.

PR #279, opened 2021-05-17
Fix K8 where conditions is not defined on pod submission

### LINKED ISSUE #278: Stack traced produced by K8 adapter
Screen Shot 2021-05-17 at 11 07 56 AM

If I run the pod YAML manually things seem to work just fine.

treydock:
Fixes #278

### DIFF
--- a/lib/ood_core/job/adapters/kubernetes/helper.rb
+++ b/lib/ood_core/job/adapters/kubernetes/helper.rb
@@ -249,11 +249,20 @@ class OodCore::Job::Adapters::Kubernetes::Helper
 
   def pod_status_from_json(json_data)
     phase = json_data.dig(:status, :phase)
+    conditions = json_data.dig(:status, :conditions)
+    unschedulable = false
+    unless conditions.nil?
+      unschedulable = conditions.any? { |c| c.dig(:reason) == "Unschedulable" }
+    end
     state = case phase
             when "Running"
               "running"
             when "Pending"
-              "queued"
+              if unschedulable
+                "queued_held"
+              else
+                "queued"
+              end
             when "Failed"
               "suspended"
             when "Succeeded"
--- a/lib/ood_core/job/adapters/kubernetes/helper.rb
+++ b/lib/ood_core/job/adapters/kubernetes/helper.rb
@@ -250,10 +250,7 @@ class OodCore::Job::Adapters::Kubernetes::Helper
   def pod_status_from_json(json_data)
     phase = json_data.dig(:status, :phase)
     conditions = json_data.dig(:status, :conditions)
-    unschedulable = false
-    unless conditions.nil?
-      unschedulable = conditions.any? { |c| c.dig(:reason) == "Unschedulable" }
-    end
+    unschedulable = conditions.to_a.any? { |c| c.dig(:reason) == "Unschedulable" }
     state = case phase
             when "Running"
               "running"
--- a/lib/ood_core/job/adapters/kubernetes/helper.rb
+++ b/lib/ood_core/job/adapters/kubernetes/helper.rb
@@ -235,6 +235,8 @@ class OodCore::Job::Adapters::Kubernetes::Helper
     if start.nil?
       # the pod is in some pending state limbo
       conditions = status.dig(:conditions)
+      return nil if conditions.nil?
+      return nil if conditions.size == 0
       # best guess to start time is just the first condition's
       # transition time
       str = conditions[0].dig(:lastTransitionTime)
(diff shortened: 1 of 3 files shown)


johrstrom:
I don't know if this is ready. Obviously a fix is needed, but I'd just like to see what the behaviour is for the user or what kubernetes responds with.

treydock:
When I submit pod that can never run, like a GPU pod when no GPUs exist, the pod status stays `Pending` with this in YAML:

```yaml
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2021-05-26T17:48:38Z"
    message: '0/3 nodes are available: 1 Insufficient nvidia.com/gpu, 1 node(s) didn''t
      match Pod''s node affinity/selector, 1 node(s) had taint {node-role.kubernetes.io/master:
      }, that the pod didn''t tolerate.'
    reason: Unschedulable
    status: "False"
    type: PodScheduled
  phase: Pending
  qosClass: Guaranteed
```

This is what I get when I try and submit the invalid pod.

```
#<OodCore::Job::Adapters::Kubernetes::Helper::K8sDataError: unable to read data correctly from json>

/opt/ood/ondemand/root/usr/share/gems/2.7/ondemand/2.0.9/gems/ood_core-0.17.0/lib/ood_core/job/adapters/kubernetes/helper.rb:162:in `rescue in pod_info_from_json'
/opt/ood/ondemand/root/usr/share/gems/2.7/ondemand/2.0.9/gems/ood_core-0.17.0/lib/ood_core/job/adapters/kubernetes/helper.rb:148:in `pod_info_from_json'
/opt/ood/ondemand/root/usr/share/gems/2.7/ondemand/2.0.9/gems/ood_core-0.17.0/lib/ood_core/job/adapters/kubernetes/helper.rb:26:in `info_from_json'
/opt/ood/ondemand/root/usr/share/gems/2.7/ondemand/2.0.9/gems/ood_core-0.17.0/lib/ood_core/job/adapters/kubernetes/batch.rb:104:in `info'
/opt/ood/ondemand/root/usr/share/gems/2.7/ondemand/2.0.9/gems/ood_core-0.17.0/lib/ood_core/job/adapters/kubernetes.rb:148:in `info'
<SNIP>
```

The pod is submitted but has a status of `Pending` like mentioned above.  Using this pull request in a dev sandbox, I get no exception but instead get redirect to regular session status page and the pod itself just shows `Starting` forever.  I'm not sure how this should be handled.  It's pretty easy to query that the this pod can't run:

```
# kubectl get pod -n user-tdockendorf -o jsonpath={.status.conditions[0].reason} jupyter-760clk2
Unschedulable
```

I added a new commit to set the status to `queued_held` if the pod is unschedulable, not sure if there might be better approach.

treydock:
This is what a job never able to run looks like now:

Screen Shot 2021-05-26 at 2 40 23 PM

treydock:
@johrstrom Does this seem more ready now that the condition of unrunnable pod is handled?  I wasn't sure if maybe needed to expand the `Status` class to handle a new status or if using `queued_held` was an acceptable idea.

johrstrom:
I think we're on the right track. I looked at this for a bit but got side-tracked looking at the history at https://github.com/OSC/ood_core/pull/230.  I think that's the right state for this given you have no where to go?  I just want to accommodate where you're never going to get anywhere and `queued` where you're actually just waiting for something to free up.

treydock:
Correct, and the user experience and UI for unscheduleable seems pretty decent, given a red box stating there is an error.  Let me know if more is needed but as far as I can tell the queued / pending for being busy is still going to happen and if the pod can never run like if the request is invalid, then there will be queued_hold.

lib/ood_core/job/adapters/kubernetes/helper.rb
```diff
@@ -247,11 +249,20 @@ def submission_time(json_data)
 
   def pod_status_from_json(json_data)
     phase = json_data.dig(:status, :phase)
+    conditions = json_data.dig(:status, :conditions)
+    unschedulable = false
+    unless conditions.nil?
+      unschedulable = conditions.any? { |c| c.dig(:reason) == "Unschedulable" }
+    end
```
johrstrom, line 256: I think this whole block can be `conditions.to_a.any? { |c| c.dig(:reason) == "Unschedulable" }` if we just cast to an array.

treydock: I didn't realize `nil.to_a` resulted in an empty array.  That sounds newish but maybe it's always been that way.  I made the necessary adjustments.

johrstrom (approved):
Thanks, I think we're good now. I created #293 as a follow up around this submission_time bug.

PR #289, opened 2021-05-27
Allow cluster batch_connect_ssh_allow? to return nil if not defined

treydock:
Based on conversation from https://github.com/OSC/ondemand/pull/1173


johrstrom (approved):
Thanks! sorry to make you do it twice (or more).

PR #290, opened 2021-05-28
Expose pod port as environment variable

treydock:
This is useful to avoid hardcoding port numbers into batch connect job templates and can also be used to pass the port number to commands launching a container.


johrstrom (approved):
Yep!

PR #291, opened 2021-05-28
Support Kubernetes startupProbe

treydock:
This will have same idea as the `after.sh.erb` used by HPC batch connect apps, the pod won't report as `Running` until the TCP port defined for the pod is actually up.  I tried to make the class structure such that we could switch to a HTTP probe instead of TCP just by adding a new class in the future.  I opted for TCP probe over HTTP probe because didn't want to have to worry about which path for HTTP probe would respond with 200 status code which could vary greatly between batch connect apps.  Using a TCP probe seemed more universal.

### DIFF
--- a/lib/ood_core/job/adapters/kubernetes/resources.rb
+++ b/lib/ood_core/job/adapters/kubernetes/resources.rb
@@ -33,13 +33,47 @@ module OodCore::Job::Adapters::Kubernetes::Resources
     end
   end
 
+  class Probe
+    attr_accessor :initial_delay_seconds, :failure_threshold, :period_seconds
+
+    def initialize(data)
+      data ||= {}
+      @initial_delay_seconds = data[:initial_delay_seconds] || 2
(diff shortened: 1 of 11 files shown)


lib/ood_core/job/adapters/kubernetes/resources.rb
```diff
@@ -33,13 +33,47 @@ def initialize(data)
...
+    def to_h
+      {
+        initial_delay_seconds: initial_delay_seconds,
+        failure_threshold: failure_threshold,
+        period_seconds: period_seconds,
+      }
+    end
+  end
+
+  class TCPProbe < Probe
```
johrstrom, line 55: I don't think we need the subclass. I think we can just start with 1 class and modify later if required. But you're right about tcp being better than http.

treydock: I merged `Probe` into `TCPProbe` so now there is only 1 class.

PR #292, opened 2021-05-28
Default to setting KUBECONFIG=/dev/null

### LINKED ISSUE #283: Kubernetes contexts collide if used outside OnDemand
Using ondemand-test I began to have initialization issues like this:

```
$ kubectl get services jupyter-2iee6rcf-service -n user-tdockendorf
Error from server (NotFound): services "jupyter-2iee6rcf-service" not found
```
This is from within a running pod.  The issue is context used is the context I setup outside OnDemand for interacting with Kubernetes dev:

```
$ kubectl config current-context
tdockendorf@kubernetes-dev
```

Ideally any kubectl commands executed from within the containers to do initialization would also be using same context flags as OnDemand so that there is not the assumption that the default context is correct.

treydock:
Fixes #283

Added test to validate someone can still override KUBECONFIG if they define it.

### DIFF
--- a/lib/ood_core/job/adapters/kubernetes/batch.rb
+++ b/lib/ood_core/job/adapters/kubernetes/batch.rb
@@ -183,6 +183,7 @@ class OodCore::Job::Adapters::Kubernetes::Batch
       HOME: home_dir,
       GROUP: group,
       GID: run_as_group,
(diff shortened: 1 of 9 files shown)


johrstrom (approved):
:+1:

PR #295, opened 2021-06-03
Use pod creation time as one way to get submission time

### LINKED ISSUE #293: try to get k8s submission time from metadata
Looking at #279, `status.conditions` is not guarenteed to be populated which means submission_time is still off. We should also check `metadata.creationTimestamp` to see if it's available (k8s docs say it SHOULD, so not guaranteed).

treydock:
Fixes #293

### DIFF
--- a/lib/ood_core/job/adapters/kubernetes/helper.rb
+++ b/lib/ood_core/job/adapters/kubernetes/helper.rb
@@ -254,6 +254,7 @@ class OodCore::Job::Adapters::Kubernetes::Helper
       # best guess to start time is just the first condition's
       # transition time
       str = conditions[0].dig(:lastTransitionTime)
+      return nil if str.nil?
     end
 
     DateTime.parse(str).to_time.to_i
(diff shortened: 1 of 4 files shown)


lib/ood_core/job/adapters/kubernetes/helper.rb
```diff
@@ -240,15 +240,18 @@ def end_time(status, state_data)
   def submission_time(json_data)
     status = json_data.dig(:status)
     start = status.dig(:startTime)
+    creation = json_data.dig(:metadata, :creationTimestamp)
 
-    if start.nil?
+    if start.nil? && creation.nil?
```
johrstrom, line 245: Seems like precedence should be creationTimestamp -> startTime -> lastTransitionTime -> nil. I think I got it wrong in the beginning saying submission_time was start time.

Feel free to re-arrange this method to fit that precedence.

treydock: Fixed, now it's creationTimestamp -> startTime -> lastTransitionTime -> nil

Issue #227, opened 2020-12-09
k8s initialization should happen statically


johrstrom:
There are a few things in the k8s adapter that should happen only once instead of every time the object is created

Specifically all this stuff that initializes the `~/.kube/confg` to create the right context and connect the user to that context.
https://github.com/OSC/ood_core/blob/53932732103b69aa8cb4a8fdd9106a745c586c19/lib/ood_core/job/adapters/kubernetes/batch.rb#L275-L278

This is 2 or 3 commands that happen on every invocation that really only need to happen once.

johrstrom:
See lots of details in #298.

Basically we can do a lot in the hooks or in a Rails initializer. 

Here's a pros and cons list of putting it in a hook. If it's a PRO here it'd be a CON for a Rails initializer. 

PROS:
* keeps all initialization in 1 thing.

CONS:
* would duplicate any cluster `allow?` configurations in the `cluster.d` files.  (i.e., you'd have to use an if block to execute them and so it wouldn't work out of the box).
* Rails seems like it'd be way more flexible and perhaps easier to manage.

treydock:
Just to add one bit of detail that some initialization must take place in hooks, ie the part about seeing up credentials since it reads a OIDC client secret as root.  The rest can be run as the user since all the information (Kubernetes API address, CA cert, etc) is not secret.

PR #298, opened 2021-07-02
Always use --context flag with K8 commands

### LINKED ISSUE #294: Kubernetes job shows as 'Completed' despite pod actually running
I was testing some changes unrelated to OnDemand and submitted a pod that is running:

```
# kubectl get pod -n user-tdockendorf
NAME               READY   STATUS    RESTARTS   AGE
jupyter-2emyppzl   1/1     Running   0          8m34s
```

But the session card for OnDemand shows `Completed` as if the pod isn't running:

Screen Shot 2021-06-03 at 9 11 11 AM

This is full pod YAML:

```yaml
apiVersion: v1
kind: Pod
metadata:
  annotations:
    cni.projectcalico.org/podIP: 192.168.255.198/32
    cni.projectcalico.org/podIPs: 192.168.255.198/32
    pod.kubernetes.io/lifetime: 01h00m00s
  creationTimestamp: "2021-06-03T13:03:29Z"
  labels:
    account: PZS0708
    app.kubernetes.io/managed-by: open-ondemand
    app.kubernetes.io/name: jupyter
    job: jupyter-2emyppzl
  name: jupyter-2emyppzl
  namespace: user-tdockendorf
  resourceVersion: "109784877"
  uid: caf345af-7cb4-4a0e-919e-d35bfa577751
spec:
  containers:
  - command:
    - /bin/bash
    - -l
    - /users/sysp/tdockendorf/ondemand/data/sys/dashboard/batch_connect/usr/tdockendorf/bc_osc_jupyter/output/50039ee4-4550-4714-97a0-306f52f6a05b/job_script_content.sh
    env:
    - name: POD_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.name
    - name: POD_NAMESPACE
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.namespace
    - name: USER
      value: tdockendorf
    - name: UID
      value: "20821"
    - name: HOME
      value: /users/sysp/tdockendorf
    - name: GROUP
      value: PZS0708
    - name: GID
      value: "5509"
    - name: NB_UID
      value: "20821"
    - name: NB_USER
      value: tdockendorf
    - name: NB_GID
      value: "5509"
    - name: CLUSTER
      value: owens
    - name: KUBECONFIG
      value: /dev/null
    image: docker-registry.osc.edu/ondemand/ondemand-base-rhel7:0.1.0
    imagePullPolicy: IfNotPresent
    name: jupyter
    ports:
    - containerPort: 8080
      protocol: TCP
    resources:
      limits:
        cpu: "1"
        memory: 4Gi
      requests:
        cpu: "1"
        memory: 4Gi
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - all
      privileged: false
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    volumeMounts:
    - mountPath: /ood/script.sh
      name: configmap-volume
      subPath: script.sh
    - mountPath: /users/sysp/tdockendorf
      name: home
    - mountPath: /fs/project
      name: project
    - mountPath: /fs/scratch
      name: scratch
    - mountPath: /fs/ess
      name: ess
    - mountPath: /var/run/munge/munge.socket.2
      name: munge-socket
    - mountPath: /etc/slurm
      name: slurm-conf
    - mountPath: /var/lib/sss/pipes
      name: sssd-pipes
    - mountPath: /etc/sssd
      name: sssd-conf
    - mountPath: /etc/nsswitch.conf
      name: nsswitch
    - mountPath: /etc/profile.d/lmod.sh
      name: lmod-init
    - mountPath: /opt/intel
      name: intel
    - mountPath: /usr/local
      name: apps
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
      name: kube-api-access-jwtb6
      readOnly: true
  dnsPolicy: ClusterFirst
  enableServiceLinks: true
  imagePullSecrets:
  - name: private-docker-registry
  nodeName: kubeworker01-dev
  nodeSelector:
    cluster: owens
  preemptionPolicy: PreemptLowerPriority
  priority: 0
  restartPolicy: OnFailure
  schedulerName: default-scheduler
  securityContext:
    fsGroup: 5509
    runAsGroup: 5509
    runAsNonRoot: true
    runAsUser: 20821
  serviceAccount: default
  serviceAccountName: default
  terminationGracePeriodSeconds: 30
  tolerations:
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 300
  volumes:
  - configMap:
      defaultMode: 420
      name: jupyter-2emyppzl-configmap
    name: configmap-volume
  - hostPath:
      path: /users/sysp/tdockendorf
      type: Directory
    name: home
  - hostPath:
      path: /fs/project
      type: Directory
    name: project
  - hostPath:
      path: /fs/scratch
      type: Directory
    name: scratch
  - hostPath:
      path: /fs/ess
      type: Directory
    name: ess
  - hostPath:
      path: /var/run/munge/munge.socket.2
      type: Socket
    name: munge-socket
  - hostPath:
      path: /etc/slurm
      type: Directory
    name: slurm-conf
  - hostPath:
      path: /var/lib/sss/pipes
      type: Directory
    name: sssd-pipes
  - hostPath:
      path: /etc/sssd
      type: Directory
    name: sssd-conf
  - hostPath:
      path: /etc/nsswitch.conf
      type: File
    name: nsswitch
  - hostPath:
      path: /apps/owens/lmod/lmod.sh
      type: File
    name: lmod-init
  - hostPath:
      path: /nfsroot/owens/opt/intel
      type: Directory
    name: intel
  - hostPath:
      path: /apps/owens
      type: Directory
    name: apps
  - name: kube-api-access-jwtb6
    projected:
      defaultMode: 420
      sources:
      - serviceAccountToken:
          expirationSeconds: 3607
          path: token
      - configMap:
          items:
          - key: ca.crt
            path: ca.crt
          name: kube-root-ca.crt
      - downwardAPI:
          items:
          - fieldRef:
              apiVersion: v1
              fieldPath: metadata.namespace
            path: namespace
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2021-06-03T13:03:30Z"
    status: "True"
    type: Initialized
  - lastProbeTime: null
    lastTransitionTime: "2021-06-03T13:05:49Z"
    status: "True"
    type: Ready
  - lastProbeTime: null
    lastTransitionTime: "2021-06-03T13:05:49Z"
    status: "True"
    type: ContainersReady
  - lastProbeTime: null
    lastTransitionTime: "2021-06-03T13:03:30Z"
    status: "True"
    type: PodScheduled
  containerStatuses:
  - containerID: containerd://e80028269e7de795fa66da861933ed73ceeb5f0a21c855f9fee3ba96c19041a2
    image: docker-registry.osc.edu/ondemand/ondemand-base-rhel7:0.1.0
    imageID: docker-registry.osc.edu/ondemand/ondemand-base-rhel7@sha256:de12a66d74578b15484954560ed0e5cd565ac60b1aabedd4dcd4989132072dd7
    lastState: {}
    name: jupyter
    ready: true
    restartCount: 0
    started: true
    state:
      running:
        startedAt: "2021-06-03T13:05:48Z"
  hostIP: 192.148.247.227
  phase: Running
  podIP: 192.168.255.198
  podIPs:
  - ip: 192.168.255.198
  qosClass: Guaranteed
  startTime: "2021-06-03T13:03:30Z"
```

treydock:
Do not use ood_core to modify KUBECONFIG
Fixes #294

The idea here is do not set cluster or context inside ood_core, just have the context and cluster exist via hooks and then always reference the necessary context with every CLI call to kubectl.  This ensures mutli cluster environments wont' be stepping on each other by making conflicting changes to the set-context and set-cluster.

### DIFF
--- a/lib/ood_core/job/adapters/kubernetes/batch.rb
+++ b/lib/ood_core/job/adapters/kubernetes/batch.rb
@@ -263,7 +263,11 @@ class OodCore::Job::Adapters::Kubernetes::Batch
   end
 
   def base_cmd
-    "#{bin} --kubeconfig=#{config_file} --context=#{context}"
+    "#{kubectl_cmd} --context=#{context}"
+  end
+
+  def kubectl_cmd
+    "#{bin} --kubeconfig=#{config_file}"
   end
 
   def all_pods_to_info(data)
@@ -291,6 +295,7 @@ class OodCore::Job::Adapters::Kubernetes::Batch
   end
 
   def make_kubectl_config(config)
+    set_cluster(config.fetch(:server, default_server).to_h.symbolize_keys)
     configure_auth(config.fetch(:auth, default_auth).to_h.symbolize_keys)
   end
 
@@ -301,6 +306,8 @@ class OodCore::Job::Adapters::Kubernetes::Batch
     case type
     when 'gke'
       set_gke_config(auth)
+    when 'oidc'
+      set_context
     end
   end
 
@@ -334,6 +341,25 @@ class OodCore::Job::Adapters::Kubernetes::Batch
     cmd = "gcloud container clusters get-credentials #{locale} #{cluster}"
     env = { 'KUBECONFIG' => config_file }
     call(cmd, env)
+  end
+
+  def set_context
+    cmd = "#{kubectl_cmd} config set-context #{cluster}"
+    cmd << " --cluster=#{cluster} --namespace=#{namespace}"
+    cmd << " --user=#{k8s_username}"
+
+    call(cmd)
+  end
+
+  def set_cluster(config)
+    server = config.fetch(:endpoint)
+    cert = config.fetch(:cert_authority_file, nil)
+
+    cmd = "#{kubectl_cmd} config set-cluster #{cluster}"
+    cmd << " --server=#{server}"
+    cmd << " --certificate-authority=#{cert}" unless cert.nil?
+
+    call(cmd)
   end
 
   def call(cmd = '', env: {}, stdin: nil)
--- a/lib/ood_core/job/adapters/kubernetes/batch.rb
+++ b/lib/ood_core/job/adapters/kubernetes/batch.rb
@@ -12,7 +12,7 @@ class OodCore::Job::Adapters::Kubernetes::Batch
   class NotFoundError < StandardError; end
 
   attr_reader :config_file, :bin, :cluster, :mounts
-  attr_reader :all_namespaces, :using_context, :helper
+  attr_reader :all_namespaces, :helper
   attr_reader :username_prefix, :namespace_prefix
 
   def initialize(options = {})
@@ -26,7 +26,6 @@ class OodCore::Job::Adapters::Kubernetes::Batch
     @username_prefix = options.fetch(:username_prefix, '')
     @namespace_prefix = options.fetch(:namespace_prefix, '')
 
-    @using_context = false
     @helper = OodCore::Job::Adapters::Kubernetes::Helper.new
 
     begin
@@ -264,9 +263,7 @@ class OodCore::Job::Adapters::Kubernetes::Batch
   end
 
   def base_cmd
-    base = "#{bin} --kubeconfig=#{config_file}"
-    base << " --context=#{context}" if using_context
-    base
+    "#{bin} --kubeconfig=#{config_file} --context=#{context}"
   end
 
   def all_pods_to_info(data)
@@ -294,7 +291,6 @@ class OodCore::Job::Adapters::Kubernetes::Batch
   end
 
   def make_kubectl_config(config)
-    set_cluster(config.fetch(:server, default_server).to_h.symbolize_keys)
     configure_auth(config.fetch(:auth, default_auth).to_h.symbolize_keys)
   end
 
@@ -305,13 +301,7 @@ class OodCore::Job::Adapters::Kubernetes::Batch
     case type
     when 'gke'
       set_gke_config(auth)
-    when 'oidc'
-      set_context
     end
-  end
-
-  def use_context
-    @using_context = true
   end
 
   def managed?(type)
@@ -344,26 +334,6 @@ class OodCore::Job::Adapters::Kubernetes::Batch
     cmd = "gcloud container clusters get-credentials #{locale} #{cluster}"
     env = { 'KUBECONFIG' => config_file }
     call(cmd, env)
-  end
-
-  def set_context
-    cmd = "#{base_cmd} config set-context #{cluster}"
-    cmd << " --cluster=#{cluster} --namespace=#{namespace}"
-    cmd << " --user=#{k8s_username}"
-
-    call(cmd)
-    use_context
-  end
-
-  def set_cluster(config)
-    server = config.fetch(:endpoint)
-    cert = config.fetch(:cert_authority_file, nil)
-
-    cmd = "#{base_cmd} config set-cluster #{cluster}"
-    cmd << " --server=#{server}"
-    cmd << " --certificate-authority=#{cert}" unless cert.nil?
-
-    call(cmd)
   end
 
   def call(cmd = '', env: {}, stdin: nil)
--- a/lib/ood_core/job/adapters/kubernetes/batch.rb
+++ b/lib/ood_core/job/adapters/kubernetes/batch.rb
@@ -344,7 +344,7 @@ class OodCore::Job::Adapters::Kubernetes::Batch
   end
 
   def set_context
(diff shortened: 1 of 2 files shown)


treydock:
Tested this by logging into ondemand.osc.edu, using K8 app , then logging into ondemand-test.osc.edu and launch another K8 app and now they are both running and second one doesn't show as incorrect completed since they are not relying on modifications to KUBECONFIG by ood_core but rather the defined context set by each environment's hook.

johrstrom:
Forcing a set known context is likely the way to go, I just need a second to think about how it would work with GKE. 

When you tested, did you wipe your ~/.kube/config?  We'd need to be sure there's not something left over from a previous initialization. I mean we'd want to be sure we don't rely on something that was created by the production ood_core

I guess semi-related is #227

treydock:
So this will have to be paired with a change to the OnDemand hooks, because we don't `set-cluster` there.  We'd also have to move the `set-context` there too.  I think we could possibly leave those items in here as long as we always pass `--context` flag which is the main thing to implement.

treydock:
I think keeping `set-cluster` and `set-context` produces the smallest change, but if it's preferred I can revert latest commit and move that logic into ondemand hooks. Just have to update configs to move some cluster.d config options into hooks since they would no longer be needed for ood_core.

johrstrom:
We can also create an initializer in OnDemand that does something like this to set the KUBECONFIG for each cluster that's defined. That way we don't have to rely on shell scripting and so on, we can do it inside Ruby with whatever settings are in the `cluster.d` file.

```
if k8s cluster is defined
  k8_clusters.each do |cluster|
    OodCore::Job::Adapters::Kubernetes::Batch.initialize_cluster!(cluster)
  end
end
```

treydock:
I think that can work.  That would just change a lot of code because right now the config is initialized differently from that, so to initialize the code via class method would have to shift how things are currently pulled out. The simplest solution would be something like this (keep using instance method):

```ruby
if k8s cluster is defined
  k8_clusters.each do |cluster|
    OodCore::Job::Adapters::Kubernetes.new.batch.initialize_cluster!(cluster)
  end
end
```

Just don't automatically build the configs during initialization like is currently done so the above would be only way to initialize configs.

johrstrom:
Well I guess then we have initialization all over the place... Clearly we'll have to think this through. 

Forcing `--context=#{context}` on every call is probably the best way to go, I'm sure of that. Whether the context is created in Ruby or though a hook I guess we'll have to think through. 

If we did this in the hooks, cluster name would duplicate in the hook env file and the cluster.d file right?  Is that the only thing the hook would need to `kubectl set context` (it already has username and kubeconfig).  Seems like it may be simpler to duplicate in a hook than to spread initialization across a bunch of different things. Though I haven't thought through GKE or `managed` use cases yet.

lib/ood_core/job/adapters/kubernetes/batch.rb
```diff
@@ -347,19 +344,18 @@ def set_gke_credentials(auth)
   end
 
   def set_context
-    cmd = "#{base_cmd} config set-context #{cluster}"
+    cmd = "#{kubectl_cmd} config set-context #{cluster}"
```
johrstrom, line 347: Looks like we should use `#{context}` here, not cluster.  I know it's an existing bug, but may as well take care of it while we see it.

treydock: Fixed. They are same thing, the `context` function just calls `cluster`, but if that ever changed this would be necessary.

treydock:
> If we did this in the hooks, cluster name would duplicate in the hook env file and the cluster.d file right? Is that the only thing the hook would need to `kubectl set context` (it already has username and kubeconfig). Seems like it may be simpler to duplicate in a hook than to spread initialization across a bunch of different things. Though I haven't thought through GKE or `managed` use cases yet.

We would also need things like path to CA file and API host address, because we would also need to do `set-cluster` in hook if we moved the logic there.  For `set-context` the only thing I think we'd need is cluster name which gets used for context.

treydock:
Added commit to not use `--context` with GKE

treydock:
If the changes to keep behavior aren't good for this PR, I can revert that commit.

treydock:
Closing this since the code was already calling `--context` and the real issue was using same Kubernetes OnDemand cluster name for different Kubernetes environments.

PR #300, opened 2021-07-12
Only show Kubernetes pod as Running when pod is ready

### LINKED ISSUE #274: K8 shows Running in OnDemand while pod is in CrashLoopBackOff state
Pod:

```
NAME               READY   STATUS             RESTARTS   AGE
jupyter-pwu8xjm0   0/1     CrashLoopBackOff   2          39s
```

OnDemand:

Screen Shot 2021-05-16 at 4 53 45 PM

treydock:
Fixes #274

### DIFF
--- a/lib/ood_core/job/adapters/kubernetes/helper.rb
+++ b/lib/ood_core/job/adapters/kubernetes/helper.rb
@@ -265,7 +265,7 @@ class OodCore::Job::Adapters::Kubernetes::Helper
     conditions = json_data.dig(:status, :conditions)
     container_statuses = json_data.dig(:status, :containerStatuses)
     unschedulable = conditions.to_a.any? { |c| c.dig(:reason) == "Unschedulable" }
-    ready = container_statuses.to_a.any? { |s| s.dig(:ready) == true }
+    ready = !container_statuses.to_a.empty? && container_statuses.to_a.all? { |s| s.dig(:ready) == true }
     if ready
       state = "running"
     else
--- a/lib/ood_core/job/adapters/kubernetes/helper.rb
+++ b/lib/ood_core/job/adapters/kubernetes/helper.rb
@@ -150,7 +150,7 @@ class OodCore::Job::Adapters::Kubernetes::Helper
     {
       id: json_data.dig(:metadata, :name).to_s,
       job_name: name_from_metadata(json_data.dig(:metadata)),
-      status: pod_status_from_json(json_data),
+      status: OodCore::Job::Status.new(state: pod_status_from_json(json_data)),
       job_owner: job_owner_from_json(json_data, ns_prefix),
       submission_time: submission_time(json_data),
       dispatch_time: dispatch_time(json_data),
@@ -266,28 +266,23 @@ class OodCore::Job::Adapters::Kubernetes::Helper
     container_statuses = json_data.dig(:status, :containerStatuses)
     unschedulable = conditions.to_a.any? { |c| c.dig(:reason) == "Unschedulable" }
     ready = !container_statuses.to_a.empty? && container_statuses.to_a.all? { |s| s.dig(:ready) == true }
-    if ready
-      state = "running"
-    else
-      state = case phase
-          when "Pending"
-            if unschedulable
-              "queued_held"
+    return "running" if ready
+    state = case phase
+            when "Pending"
+              if unschedulable
+                "queued_held"
+              else
+                "queued"
+              end
+            when "Failed"
+              "suspended"
+            when "Succeeded"
+              "completed"
+            when "Unknown"
+              "undetermined"
             else
-              "queued"
+              "undetermined"
             end
-          when "Failed"
-            "suspended"
-          when "Succeeded"
-            "completed"
-          when "Unknown"
-            "undetermined"
-          else
-            "undetermined"
-          end
-    end
-
-    OodCore::Job::Status.new(state: state)
   end
 
   def terminated_state(status)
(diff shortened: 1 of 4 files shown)


treydock:
The diff makes much more sense with `git log -p -w` or `git diff -w`:

```diff

--- a/lib/ood_core/job/adapters/kubernetes/helper.rb
+++ b/lib/ood_core/job/adapters/kubernetes/helper.rb
@@ -263,10 +263,13 @@ class OodCore::Job::Adapters::Kubernetes::Helper
   def pod_status_from_json(json_data)
     phase = json_data.dig(:status, :phase)
     conditions = json_data.dig(:status, :conditions)
+    container_statuses = json_data.dig(:status, :containerStatuses)
     unschedulable = conditions.to_a.any? { |c| c.dig(:reason) == "Unschedulable" }
+    ready = container_statuses.to_a.any? { |s| s.dig(:ready) == true }
+    if ready
+      state = "running"
+    else
       state = case phase
-            when "Running"
-              "running"
           when "Pending"
             if unschedulable
               "queued_held"
@@ -282,6 +285,7 @@ class OodCore::Job::Adapters::Kubernetes::Helper
           else
             "undetermined"
           end
+    end
 
     OodCore::Job::Status.new(state: state)
   end
```

lib/ood_core/job/adapters/kubernetes/helper.rb
```diff
@@ -263,25 +263,29 @@ def submission_time(json_data)
...
-              else
-                "queued"
-              end
-            when "Failed"
-              "suspended"
-            when "Succeeded"
-              "completed"
-            when "Unknown"
-              "undetermined"
+    ready = container_statuses.to_a.any? { |s| s.dig(:ready) == true }
```
johrstrom, line 268: Shouldn't this be `!container_statuses.empty? && container_statuses.to_a.all? { |s| s.dig(:ready) == true }` to account for other containers (that we don't yet support but, maybe will and we'll lose this).

I'm not in love with an if block and a case statemement, but I think the only other alternative is a return statement with is also not great.

johrstrom:
Can you add tests for the cases you've found in #274.

lib/ood_core/job/adapters/kubernetes/helper.rb
treydock: So the `.all?` would only be useful in context of multiple containers, not init containers because the key for init containers is `initContainerStatuses`.  But having multiple containers is something needing to be supported so I can update that.

treydock: Made this change, also made it so the info function only returns string, so could return early when ready. Then the place needing a status object is what sets the status.

treydock:
Added test cases for issue pulling down image and also CrashLoopBackoff.  This is what CrashLoopBackoff now looks like:

Screen Shot 2021-07-12 at 11 41 23 AM

lib/ood_core/job/adapters/kubernetes/helper.rb
treydock: There were also many places where `pod_status_from_json` was getting used by helper functions and expecting a lowercase string for comparison so to ensure that actually works and is clearly going to work, moved the `OodCore::Job::Status.new` call to the place where that object type is actually needed so `pod_status_from_json` just returns a status string.

spec/fixtures/output/k8s/single_image_error_pod.json
```diff
@@ -0,0 +1,457 @@
+{
```
johrstrom, line 1: Did you manually update these files from existing files?  I think we should pull an existing json files from actual image error and crash loop pods (I'm surprised there isn't a crashloop json in here already?)

lib/ood_core/job/adapters/kubernetes/helper.rb
```diff
@@ -263,10 +263,11 @@ def submission_time(json_data)
   def pod_status_from_json(json_data)
     phase = json_data.dig(:status, :phase)
     conditions = json_data.dig(:status, :conditions)
+    container_statuses = json_data.dig(:status, :containerStatuses)
     unschedulable = conditions.to_a.any? { |c| c.dig(:reason) == "Unschedulable" }
+    ready = !container_statuses.to_a.empty? && container_statuses.to_a.all? { |s| s.dig(:ready) == true }
+    return "running" if ready
     state = case phase
```
johrstrom, line 270: I don't think we have a rubocop here yet, but I believe it would like a newline after this return statement.

spec/fixtures/output/k8s/single_image_error_pod.json
treydock: I took an existing pod JSON and modified the relevant parts, ie the `status`.  I been testing with an app that doesn't use init containers and also has a long list of env variables and mounts so didn't want to create a massive file when the only part that matters is the status.  The status in the JSON file I added was copy+pasted from the actual output with kubectl, the other parts were just copied from existing JSON files.

lib/ood_core/job/adapters/kubernetes/helper.rb
treydock: Fixed.

spec/fixtures/output/k8s/single_image_error_pod.json
treydock: Should I regenerate these new JSON files will full contents copied from kubectl output? The files will be rather large unless I generate the errors with a slimmed down app that doesn't mount as many things.

johrstrom: Yes please - I'd think it's just safer.  The size I'm not super worried about, I'd have to check the gemspec, but I don't believe test/ directory gets put into the gem (again I'd have to check to be sure).

In any case - yes I think real kubectl output is best.

treydock: Added full JSON for the 2 test cases added in this pull request.

johrstrom (approved):
LGTM, thanks

PR #303, opened 2021-07-21
Kubernetes - handle state when pod is Running but not Ready

treydock:


treydock:
The POD YAML for the test was direclty pulled from pod that got into "Undetermined" state which is incorrect state for a pod that is simply not ready yet. I forced "not ready" by never launching RStudio and instead just sleeping

johrstrom (approved):
Works for me.

PR #305, opened 2021-07-27
Better handling of supplemental groups in Kubernetes

### LINKED ISSUE #304: Support supplemental groups in k8s
We're seeing this output from `/proc/$$/self` in containers.  I suspect that the kublet is just starting the container as prescribed, i.e., with no supplemental groups. SSSD only kicks in after the container entrypoint is allocated, so we can't see files in our supplemental groups (like scratch or project directories).

Here's the output in a RStudio server container, when ran anywhere else would have a lot more ids - all my supplemental groups.
```text - 
> system('grep ^Groups /proc/$$/status')
Groups:	5515
```

This widens the attack surface, so at the least - we'll have to `groups.reject(id <= 1024)`.  On the kubernetes side we'll have to document what we do.

treydock:
Fixes #304 

Also previous behavior was wrong, can't wrap groups in quotes, they must be integers: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#podsecuritycontext-v1-core

If we don't want to assume a default then this PR could be just to remove quotes and move logic of using OodSupport to get groups to each app but seems friendlier to our community to have sane default in ood_core.

### DIFF
--- a/lib/ood_core/job/adapters/kubernetes/batch.rb
+++ b/lib/ood_core/job/adapters/kubernetes/batch.rb
@@ -177,7 +177,7 @@ class OodCore::Job::Adapters::Kubernetes::Batch
   end
 
   def supplemental_groups
-    OodSupport::User.new.groups.sort_by(&:id).map(&:id)
+    OodSupport::User.new.groups.sort_by(&:id).map(&:id).reject { |id| id < 1000 }
   end
 
   def default_env
--- a/lib/ood_core/job/adapters/kubernetes/batch.rb
+++ b/lib/ood_core/job/adapters/kubernetes/batch.rb
@@ -14,6 +14,7 @@ class OodCore::Job::Adapters::Kubernetes::Batch
   attr_reader :config_file, :bin, :cluster, :mounts
   attr_reader :all_namespaces, :using_context, :helper
   attr_reader :username_prefix, :namespace_prefix
+  attr_reader :enable_supplemental_groups
 
   def initialize(options = {})
     options = options.to_h.symbolize_keys
@@ -25,6 +26,7 @@ class OodCore::Job::Adapters::Kubernetes::Batch
     @all_namespaces = options.fetch(:all_namespaces, false)
     @username_prefix = options.fetch(:username_prefix, '')
     @namespace_prefix = options.fetch(:namespace_prefix, '')
+    @enable_supplemental_groups = options.fetch(:enable_supplemental_groups, false)
 
     @using_context = false
     @helper = OodCore::Job::Adapters::Kubernetes::Helper.new
@@ -176,8 +178,17 @@ class OodCore::Job::Adapters::Kubernetes::Batch
     Etc.getgrgid(run_as_group).name
   end
 
-  def supplemental_groups
+  def default_supplemental_groups
     OodSupport::User.new.groups.sort_by(&:id).map(&:id).reject { |id| id < 1000 }
+  end
+
+  def supplemental_groups(groups = [])
+    sgroups = []
+    if enable_supplemental_groups
+      sgroups.concat(default_supplemental_groups)
+    end
+    sgroups.concat(groups.to_a)
+    sgroups.uniq.sort
   end
 
   def default_env
@@ -195,6 +206,7 @@ class OodCore::Job::Adapters::Kubernetes::Batch
   # create an id.
   def generate_id_yml(script)
     native_data = script.native
+    native_data[:container][:supplemental_groups] = supplemental_groups(native_data[:container][:supplemental_groups])
     container = helper.container_from_native(native_data[:container], default_env)
     id = generate_id(container.name)
     configmap = helper.configmap_from_native(native_data, id, script.content)
--- a/lib/ood_core/job/adapters/kubernetes/batch.rb
+++ b/lib/ood_core/job/adapters/kubernetes/batch.rb
@@ -14,7 +14,7 @@ class OodCore::Job::Adapters::Kubernetes::Batch
   attr_reader :config_file, :bin, :cluster, :mounts
   attr_reader :all_namespaces, :using_context, :helper
   attr_reader :username_prefix, :namespace_prefix
-  attr_reader :enable_supplemental_groups
+  attr_reader :auto_supplemental_groups
 
   def initialize(options = {})
     options = options.to_h.symbolize_keys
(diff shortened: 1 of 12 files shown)


lib/ood_core/job/adapters/kubernetes/batch.rb
```diff
@@ -176,6 +176,10 @@ def group
     Etc.getgrgid(run_as_group).name
   end
 
+  def supplemental_groups
+    OodSupport::User.new.groups.sort_by(&:id).map(&:id)
```
johrstrom, line 180: We should definitely `map(&:id).reject { |id| id < 1024 }`.  If a system has a minimum of 500 UID, does the same apply to group ids?

treydock: So I did `id < 1000` because at least at OSC and the default for RHEL7+ is to do user UID/GID as 1000+.  The default for RHEL6 was 500+ but I believe during migration to OSCUS, we moved all UIDs/GIDs below 1000.

Really the minimum UID/GID is really not enforced much by anything.  There are places that configure like `useradd` and `groupadd` commands that when UID/GID is not specified, it picks from a range depending on if the record created is flagged as system account. For RHEL7+ if you tell `useradd` or `groupadd` to create a system record it will be less than 1000.  For OSC and others using LDAP, it really just comes down to standards of what is created in LDAP, since I could easily create a group with GID 200 and it is treated not differently than group with GID 2000.

johrstrom:
I think I'm OK with it, but it should be opt-in, set at the cluster level. Seems like we'd also want to filter/reject the container configs groups too, so there may be some synergy there to put all this logic into a method instead of in the template.

treydock:
> Seems like we'd also want to filter/reject the container configs groups too, so there may be some synergy there to put all this logic into a method instead of in the template.

Not sure I follow what you mean by `container configs groups`?  The current implementation is either auto groups list based on OodSupport groups, or user defined list from submit.yml.erb or no supplemental groups.

johrstrom:
Yea I interchanged `container config groups` with `user defined list`. 

But what I really mean is, let's put all this group choice login into a method and guarantee an empty array - that way the template can just that method directly without an if/else block. 

Looks like we don't need to reformat the YAML array in the else block.
```
2.7.1 :018 > [1,2,3].to_s
 => "[1, 2, 3]" 
```

treydock:
I added a commit, so that now you have to set `enable_supplemental_groups` in cluster config to get default supplemental groups from OodSupport.  The groups from OodSupport (if enabled) are merged with submit.yml defined supplemental groups.  To keep ERB simpler I opted that if no supplemental groups are defined, just don't include that key.

treydock:
Also the reason I sort the groups list is so that our policy engine can compare the group list from pod YAML against valid group list in configmap from LDAP. If both values are not sorted, there's no real way to compare the two for equality.

lib/ood_core/job/adapters/kubernetes/templates/pod.yml.erb
```diff
@@ -20,12 +20,10 @@ spec:
...
     runAsNonRoot: true
-    <%- if spec.container.supplemental_groups.empty? -%>
-    supplementalGroups: []
-    <%- else -%>
+    <%- unless spec.container.supplemental_groups.empty? -%>
     supplementalGroups:
     <%- spec.container.supplemental_groups.each do |supplemental_group| -%>
-    - "<%= supplemental_group %>"
+    - <%= supplemental_group %>
     <%- end -%>
```
johrstrom, line 27: Seems like the template should just be this with no blocks. We're guaranteeing it's empty right?  I'm just concerned with how unmanageable this template is becoming with switches everywhere.

```ruby
supplementalGroups: <%= spec.container.supplemental_groups %>
```
We have test cases where it's empty, but we'd need some test cases around it being populated too. (the test cases too could use a little re-factoring given all the options/ways we initialize the subject).

treydock: There already is one test case with `enable_supplemental_groups` cluster option being enabled so that covers the default groups getting added.  I added another test case around merging submit.yml group list with default group list.

Want me to refactor tests here or just get tests covering new behavior and refactor in different pull request?

treydock: Also added commit to simplify the template per your suggestion.

johrstrom: just get tests covering new behavior. There's no need to refactor in this PR, I'm just complaining. I just want to be sure that we actually do _guarantee_ an empty array every time.

treydock: Then that should be covered, the majority of YAML file tests verify default behavior is to do empty supplemental groups, then one test verifies we populate defaults and another test verifies we populate defaults and merge with submit.yml groups.

spec/job/adapters/kubernetes/batch_spec.rb
```diff
@@ -414,6 +417,71 @@ def build_script(opts = {})
...
+      # make sure template get's passed into command correctly
+      allow(Open3).to receive(:capture3).with(
+        {},
+        "/usr/bin/kubectl --kubeconfig=#{ENV['HOME']}/.kube/config " \
+        "--namespace=testuser -o json create -f -",
+        stdin_data: pod_yml_extra_groups.to_s
+      ).and_return(['', '', success])
+
+      @basic_batch.submit(script)
+    end
```
johrstrom, line 483: This is what I meant by the need to refactoring here. There's a whole lot of test case here and what we're testing is a very slight difference from all the other test cases.

johrstrom: Nothing to do here on this PR - just noting it for #306.

lib/ood_core/job/adapters/kubernetes/batch.rb
```diff
@@ -25,6 +26,7 @@ def initialize(options = {})
     @all_namespaces = options.fetch(:all_namespaces, false)
     @username_prefix = options.fetch(:username_prefix, '')
     @namespace_prefix = options.fetch(:namespace_prefix, '')
+    @enable_supplemental_groups = options.fetch(:enable_supplemental_groups, false)
```
johrstrom, line 29: OK - last thing. Is this the right name?  Sounds like it's disabling it altogether. Shouldn't it be `auto_populate_groups` or `auto_supplemental_groups`? I mean I think you can infer supplemental groups from the word _groups_, but you never know what people may assume.

treydock: I renamed to `auto_supplemental_groups` per suggestion.

treydock:
Let me know if any more changes needed. I have necessary Kyverno policy changes staged to support this so we can verify the `supplimentalGroups` is a list of valid GIDs for a given user based on LDAP.

treydock:
Also some changes to automation needed for this, at least for Puppet, but once this is tagged and PR opened with OnDemand, just add automation label there.

johrstrom (approved):
Yea we're good.

PR #307, opened 2021-07-29
Add Github action to publish the gem

### LINKED ISSUE #225: publish gem to rubygems on tag
Create a github action to publish gem to rubygems for every tag.

### LINKED ISSUE #285: publish to ruby gems on tags
We manually create tags and build the gem then publish to rubygems. This ticket is to build automation around publishing to rubygems through a service account when tags are created.

treydock:
Fixes #285 
Fixes #225 

Requires someone add `OSC_RUBYGEMS_API_KEY` secret to OSC organization.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,23 @@
+name: Publish Gem
+
+on:
+  push:
+    branches:
+      - "*"
+    tags:
+      - "v*"
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+
+      - name: Publish Gem
+        uses: cadwallion/publish-rubygems-action@master
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+          RUBYGEMS_API_KEY: ${{secrets.OSC_RUBYGEMS_API_KEY}}
+          RELEASE_COMMAND: rake release
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -33,7 +33,6 @@ jobs:
           mkdir -p ~/.gem
           cat << EOF > ~/.gem/credentials
           ---
-          :github: Bearer ${{secrets.GITHUB_TOKEN}}
           :rubygems_api_key: ${{secrets.OSC_RUBYGEMS_API_KEY}}
           EOF
           chmod 0600 ~/.gem/credentials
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+
+      - name: Publish Gem
+        uses: cadwallion/publish-rubygems-action@master
```
johrstrom, line 17: I don't want to user a master version of a 3rd party GH action, especially when credentials are involved. The owner is a GH employee so they must have some special way of publishing because they don't have an `action.yml`.

I guess it sounds like I think other folks' GH actions are suspect.  Maybe I do. It's easy for us to _forget_ that we have this dependency when someone could be just harvesting our gem api keys.

It's so simple I'd rather just do it ourselves so we can never have to think about the security implications (well... depending on where we place the files in the action!)

.github/workflows/publish.yml
```diff
@@ -0,0 +1,21 @@
...
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+          RUBYGEMS_API_KEY: ${{secrets.OSC_RUBYGEMS_API_KEY}}
+          RELEASE_COMMAND: rake release
```
johrstrom: You can reference `OSC_ROBOT_RUBYGEMS_TOKEN`.   

We don't seem to have an organization wide github token for the robotic accounts. But I don't think we need it here anyway? We're making tags - it's only pushing to rubygems.

Maybe all we need is:
```
VERSION=$(figure out version)
rake build
gem push "pkg/ood_core-$VERSION.gem"
```

.github/workflows/publish.yml
treydock: Updated to essentially do what the previous action did but do it all directly in the action from this repo.

.github/workflows/publish.yml
johrstrom: I left this comment just as you pushed, but I think it applies anyhow. I think now it may actually be easier to just create the token instead of parsing the GITHUB_REF?

treydock: Removed need for GITHUB_TOKEN and used `rake build` and `gem push` instead of `rake release`

johrstrom (approved):
Lol I wanted this so bad I created 2 tickets for it. Let's tag some gems! (after we update the CHAGNELOG and bump the version)

treydock:
@johrstrom Before you create tags, don't forgot to define `OSC_RUBYGEMS_API_KEY` secret for OSC organization.

johrstrom:
Oh yea, I created it before I'd dropped that comment, but put it in the ondemand repository not the organization wide!

PR #311, opened 2021-07-29
dont save gh action files

johrstrom:
dont save gh action files. There's no security risk or anything, they just don't need to be distributed.


Oglopf (approved):
LGTM.

PR #314, opened 2021-08-18
Do not wait for K8 resources to delete

treydock:
Fixes https://github.com/OSC/ondemand/issues/1348

### DIFF
--- a/lib/ood_core/job/adapters/kubernetes/batch.rb
+++ b/lib/ood_core/job/adapters/kubernetes/batch.rb
@@ -125,7 +125,7 @@ class OodCore::Job::Adapters::Kubernetes::Batch
       when "get"
         call_json_output('get', resource, id)
       when "delete"
-        call("#{namespaced_cmd} delete #{resource} #{id}")
+        call("#{namespaced_cmd} delete #{resource} #{id} --wait=false")
       end
     rescue NotFoundError
       {}
--- a/spec/job/adapters/kubernetes/batch_spec.rb
+++ b/spec/job/adapters/kubernetes/batch_spec.rb
@@ -840,25 +840,25 @@ EOS
       allow(Open3).to receive(:capture3).with(
         {},
         "/usr/bin/kubectl --kubeconfig=#{ENV['HOME']}/.kube/config " \
-        "--namespace=testuser delete pod test-pod-123",
+        "--namespace=testuser delete pod test-pod-123 --wait=false",
         stdin_data: ""
       ).and_return(['', '', success])
       allow(Open3).to receive(:capture3).with(
         {},
         "/usr/bin/kubectl --kubeconfig=#{ENV['HOME']}/.kube/config " \
-        "--namespace=testuser delete secret test-pod-123-secret",
+        "--namespace=testuser delete secret test-pod-123-secret --wait=false",
         stdin_data: ""
       ).and_return(['', '', success])
       allow(Open3).to receive(:capture3).with(
         {},
         "/usr/bin/kubectl --kubeconfig=#{ENV['HOME']}/.kube/config " \
-        "--namespace=testuser delete service test-pod-123-service",
+        "--namespace=testuser delete service test-pod-123-service --wait=false",
         stdin_data: ""
       ).and_return(['', '', success])
       allow(Open3).to receive(:capture3).with(
         {},
         "/usr/bin/kubectl --kubeconfig=#{ENV['HOME']}/.kube/config " \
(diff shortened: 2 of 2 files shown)


treydock:
Just tested this in my dev dashboard and the delete took a few seconds and my job's card was deleted and saw this a few seconds later on terminal:

```
# kubectl get pod -n user-tdockendorf
NAME               READY   STATUS        RESTARTS   AGE
jupyter-6nbmjdk8   1/1     Terminating   0          65s
```

So this worked to have OnDemand execute delete command and similar behavior to SLURM now.

johrstrom:
This is fine, but I wonder what Info.status we map to when a pod is in `Terminating` phase.

I was typing that as you left that comment. So it gets mapped to completed immediately?

treydock:
What I experienced was the card that would show the status was deleted from the page before the page could have status updated.  If the card was to get an update, it would show as `Running` because the actual status fields don't get updated with anything related to `Terminating`. What I do notice is this:

```yaml
  metadata:
    annotations:
      cni.projectcalico.org/podIP: 192.168.32.235/32
      cni.projectcalico.org/podIPs: 192.168.32.235/32
      pod.kubernetes.io/lifetime: 01h00m00s
    creationTimestamp: "2021-08-18T20:26:56Z"
    deletionGracePeriodSeconds: 30
    deletionTimestamp: "2021-08-18T20:28:00Z"
```

So the `deletionTimestamp` exists for a deleted pod.  Here are status values that was from a `Terminating` pod

```yaml
  status:
    conditions:
    - lastProbeTime: null
      lastTransitionTime: "2021-08-18T20:26:57Z"
      status: "True"
      type: Initialized
    - lastProbeTime: null
      lastTransitionTime: "2021-08-18T20:27:12Z"
      status: "True"
      type: Ready
    - lastProbeTime: null
      lastTransitionTime: "2021-08-18T20:27:12Z"
      status: "True"
      type: ContainersReady
    - lastProbeTime: null
      lastTransitionTime: "2021-08-18T20:26:57Z"
      status: "True"
      type: PodScheduled
    containerStatuses:
    - containerID: containerd://b4289f05f7554c7470d31e1a3b65f3424ec77f8115aab24e36403f6cac49dec2
      image: docker-registry.osc.edu/ondemand/ondemand-base-rhel7:0.2.0
      imageID: docker-registry.osc.edu/ondemand/ondemand-base-rhel7@sha256:5d3ffb82f3c512e476948aa0c868b42d840f24bd7b335a20b337a95628168583
      lastState: {}
      name: jupyter
      ready: true
      restartCount: 0
      started: true
      state:
        running:
          startedAt: "2021-08-18T20:26:59Z"
    hostIP: 192.148.247.170
    phase: Running
    podIP: 192.168.32.235
    podIPs:
    - ip: 192.168.32.235
    qosClass: Guaranteed
    startTime: "2021-08-18T20:26:57Z"
```

treydock:
I think we'd have to teach ood_core what a deleted pod looks like by updating the status helper function. Want me to do that in this PR or leave as is? As best I could tell from testing the card is simply deleted so no place to show status.

treydock:
Actually looking at the code I see `end_time` function but the JSON it tries to pull out for completed won't ever get hit because pods won't ever get into `completed` state and that JSON path I didn't see as existing when I deleted a pod.  All I saw change was `deletionTimestamp`

https://github.com/OSC/ood_core/blob/254e51279e6c1c7c79c4fd4afefc61e19a1d7ff7/lib/ood_core/job/adapters/kubernetes/helper.rb#L228-L239

johrstrom:
Yea it appears that the dashboard just deletes the session and returns - it doesn't say, delete then confirm to see if it's actually deleted. 

If the adapter returns 0 status (or doesn't raise an error) the dashboard takes it on face value that everything's done and cleans everything up.

I think we're good on this - but I'm going to sleep on it. It seems like there are edge cases where we'd delete something and move on but it actually still exists.  I guess the pod reaper will take care of it eventually? Or there are other administrative guardrails in place for pods that need to be force deleted?

treydock:
Yes, the job-pod-reaper would cleanup anything left behind if the delete failed for whatever reason and if job-pod-reaper is unable to delete things that produces error metrics in Prometheus and we have alerts for that.

treydock:
@johrstrom How is this looking to you? Wasn't sure if we wanted to try to get this pushed out before classes start next week, or maybe just shortly after they start?

johrstrom (approved):
Yea we do - hoping to release today I just need to get through some classes.

PR #318, opened 2021-08-23
Allow K8 CPU/memory request/limit to be different values

### LINKED ISSUE #316: Allow K8 CPU/memory request/limit to be different values
If we wanted to allocate sub-core allocations in Kubernetes, might be useful to set a request for CPU to 0.5 but allow the pod to use up to 1 core.  This could be done in backwards compatible manner, example:

current:
```yaml
script:
  native:
    container:
      cpu: "1"
      memory: "4000Mi"
```

new keys:
```yaml
script:
  native:
    container:
      cpu_request: "0.5"
      cpu_limit: "1"
      memory_request: "2000Mi"
      memory_limit: "4000Mi"
```

If the new keys are specified they will override `cpu` and `memory` since the new keys are more specific.

treydock:
Fixes #316

If the new `_limit` or `_request` are not specified the old `cpu` and `memory` values are used.


lib/ood_core/job/adapters/kubernetes/helper.rb
```diff
@@ -50,7 +50,11 @@ def container_from_native(container, default_env)
       port: container[:port],
       env: default_env.merge(env),
       memory: container[:memory],
+      memory_limit: container[:memory_limit],
+      memory_request: container[:memory_request],
```
johrstrom, line 54: Maybe we should just get rid of `memory` and `cpu` altogether and put all the OR logic here in the builder?

```ruby
memory_limit: container[:memory_limit] || container[:memory]
```

treydock: That's cleaner, I made the necessary adjustments.

spec/job/adapters/kubernetes/helper_spec.rb
```diff
@@ -342,8 +344,10 @@
             HOME: '/over/here',
             UID: 1000,
           },
-          memory: '12Gi',
-          cpu: '6',
+          memory_limit: '12Gi',
+          memory_request: '12Gi',
+          cpu_limit: '6',
+          cpu_request: '6',
```
johrstrom, line 350: Sorry, did that request make it not backward compatible?  I meant just to get rid of `memory` and `cpu` just internally - I think we can still expose & interpret them, at least for now for comparability, but also just for convenience.

treydock: Those changes are to `Kubernetes::Resources::Container.new` which no longer accepts `cpu` or `memory`.  The actual container YAML can still have those values:

https://github.com/OSC/ood_core/blob/908036a5c4b8cd88afbc587179cff9d7f5e9d468/lib/ood_core/job/adapters/kubernetes/helper.rb#L52-L55

The `Container` class is the internal change which is what the tests had to be updated to account for.  The batch spec still validate old behavior:

https://github.com/OSC/ood_core/blob/908036a5c4b8cd88afbc587179cff9d7f5e9d468/spec/job/adapters/kubernetes/batch_spec.rb#L374-L375

spec/job/adapters/kubernetes/helper_spec.rb
```diff
@@ -316,8 +316,10 @@
         env: {
           'HOME' => '/over/here',
         },
-        memory: '12Gi',
-        cpu: '6',
+        memory_limit: '12Gi',
+        memory_request: '12Gi',
+        cpu_limit: '6',
+        cpu_request: '6',
```
johrstrom, line 322: Got it. Yea looking this over again `ctr_hash` is supposed to mimic what we read out of a YAML.

Can we reset this so we ensure backward comparability/convienience. Then maybe add some test cases similar to `correctly parses container with no restart_policy` where we delete and or modify what's in `ctr_hash` appropriately?

I guess I'm interested in 
* when they're all 3 nil
* when they're all 3 set 
* when they're all 3 different

treydock: So backwards compatibility is there, the `batch_spec` tests prove that. What's not backwards compatible is what we pass to the `Container` class, but that doesn't change the YAML.  I think the issue here is the actual testing is testing the wrong thing or in the wrong way.  The actual YAML keys of `cpu` and `memory` still work so apps like what we have for Jupyter and RStudio won't have to change for this PR.

treydock: For backwards compatibility, I think I found a way to do it in helper tests:

https://github.com/OSC/ood_core/pull/318/commits/bdc3c591504e8ac73c0855774af0f1d2664173be

I will work on the other interesting tests.

treydock: Added some extra test cases, so now the majority of helper tests look at `cpu` and `memory` and the extra tests validate when no limits of any kind are set and also validate when `cpu` and `memory` differ from the `_limit` keys.

spec/job/adapters/kubernetes/helper_spec.rb
```diff
@@ -481,12 +495,68 @@
...
+
+    it "when cpu, memory and limits/requests are defined" do
       # expected defaults
       ctr_hash[:env] = {}
       ctr_hash[:command] = []
       ctr_hash.delete(:port)
       ctr_hash[:memory] = '4Gi'
+      ctr_hash[:memory_limit] = '8Gi'
+      ctr_hash[:memory_request] = '4Gi'
       ctr_hash[:cpu] = '1'
```
johrstrom, line 557: Great! Can we change 1 last thing here with, setting :memory and :cpu to some number that _isn't_ the limit or request here. I just want to be sure the output is _request and not the original.

```ruby
      ctr_hash[:memory] = '9000Gi'
      ctr_hash[:cpu] = '9000'
```

treydock: Done, made the requested changes

PR #328, opened 2021-10-14
fix --all-namespaces commands to have -o json before the verb

johrstrom:
I noticed in #324 that because we have this toggle here, most calls are`-o json <VERB>` but when we use --all-namespaces it's `<VERB> -o json`.

It's not a functionality thing, just a uniformity thing, all these calls should at least look the same if they're not coming from the same code paths (how to fix so many `call` functions here... is a larger problem).


treydock:
My only suggestion is one I've seen some other projects document is when possible put all positional arguments first followed by all flags. So do `kubectl <verb> <object> -o json --all-namespaces` kind of thing.  Functionally they are all the same.

PR #743, opened 2022-02-06
Add a linux_systemd job adapter

plazonic:
This is a job adapter that uses systemd - in particular systemd-run and systemctl show. It is heavily based on linux_host adapter with minimal changes.

Its configuration is even simpler than linux_host, e.g.:

v2:
  metadata:
    title: My Login node
  login:
    host: login1
  job:
    cluster: login1
    bin: "/usr/bin"
    adapter: linux_systemd
    submit_host: login1
    ssh_hosts:
    - login1
    site_timeout: 2678400
    strict_host_checking: false

As far as how it works - systemd-run is used to create a transient systemd user unit - this bit from linux_systemd/templates/script_wrapper.erb illustrates well what can/is set:

systemd-run --user -r --no-block --unit=<%= session_name %> -p RuntimeMaxSec=<%= script_timeout %> \
        -p ExecStartPre="$systemd_service_tmp_file_pre" -p ExecStartPost="$systemd_service_tmp_file_post" \
        -p StandardOutput="file:<%= output_path %>" -p StandardError="file:<%= error_path %>" \
        -p Description="<%= job_name %>" "$systemd_service_tmp_file"

This creates a user unit called session_name (starting with "ondemand-"). Note that systemd-run takes care of much of the required functionality - timeout, pre/post scripts (for emails), description ... Programs started under such units live in their own systemd slice/cgroup and therefore can be managed/stopped/limited.

To check on a "job" status "systemctl --user show -t service --state=running ondemand-*" is parsed.

One caution - some of the systemd features might require a new enough systemd version - in particular StandardOutput=file:/..... This was tested to work on RHEL8 with desktop app.

This code clearly requires more work - in particular it is missing testing/specs entirely (sorry, don't know how :( ) but maybe you and others could find it useful as is. More testing and maybe more scripts for reacting to job failures and sending emails are needed. Per job limits (memory/cpu) could be added fairly easily - we didn't need them for our desktop app - we are relying on systemd user cgroup based limits.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201775990272808) by [Unito](https://www.unito.io)


johrstrom:
Hi, thanks for this. I've been thinking about a localhost adapater - and this may suite the bill just fine.

It does need work - even the linux host adapter does too.  I need time to look this over. It may go into a different branch while we work somethings out. 


I'm not familiar with running systemd in usersspace (`systemd-run --user`).  I take it this doesn't need privilege to use? Any old user can start/stop a unit file like this?

treydock:
I'm fairly certain RHEL7 does not support user units, but looks like RHEL8 does support this.  https://access.redhat.com/solutions/5101061

lib/ood_core/job/adapters/linux_systemd/templates/script_wrapper.erb.sh
```diff
@@ -0,0 +1,57 @@
...
+
+echo ""
+echo "HOSTNAME:$hostname"
+
+# we need this user to be enabled for lingering or else the newly started
+# service will end as soon as the ssh session starting has exited
+loginctl enable-linger
+
+# Put the script into a temp file on localhost
+systemd_service_tmp_file_base=$(mktemp -d "<%= workdir %>/XXXXXXXXXXXXX" --suffix '_systemd')
```
johrstrom, line 24: This doesn't need to be random, it's already in a staged root directory that has a UID in it like `~ondemand/data/batch_connect/dev/systemd_test/output/16c55f46-5f5e-43d8-8853-774b5a1eab7f`.

plazonic: Hello, sorry for the delay - got distracted and forgot... 

Anyway - yeah, I pretty much knew that but I figured since linux_host does something similar I should do the same here. Maybe it was needed for testing or some other reason I wasn't familiar with. Easy enough to remove/simplify and can do that if told.

johrstrom:
I'm having trouble testing this. Trying to avoid spawning a VM. I have a rockylinux:8 container with systemd but it's getting stuck on the `loginctl`. 

I may have to still spawn that VM. In any case - I'm on this and trying to replicate it.

lib/ood_core/job/adapters/linux_systemd/templates/script_wrapper.erb.sh
johrstrom: Originally they were just placed in `$HOME` so I think they needed to be random.  I think I patched it after the fact to drop them in `staged_root` for debugging purposes.

Now that you mention it, I doubt they need to be randomized either. I think making a directory like `<%=workdir %>.unit_files` is fine.

lib/ood_core/job/adapters/linux_systemd/launcher.rb
```diff
@@ -0,0 +1,247 @@
...
+        'job_name' => (script.job_name) ? script.job_name : 'LinuxHost_Adapter_Job',
+        'job_status' => event
+      }.each{
+        |key, value| bnd.local_variable_set(key, value)
+      }
+    })
+  end
+
+  def unique_session_name
+    "#{session_name_label}-#{SecureRandom.uuid}"
```
johrstrom, line 195: Can this be something shorter than however long uuid is. In the UI it's just too long for the card.

plazonic: That makes a lot of sense. Shortened to random 10 alphanumeric chars - should be good enough and hopefully shorter.

lib/ood_core/job/adapters/linux_systemd/templates/script_wrapper.erb.sh
plazonic: Done - thanks.

lib/ood_core/job/adapters/linux_systemd/launcher.rb
```diff
@@ -0,0 +1,247 @@
...
+    }.sort.join("\n")
+  end
+
+  def script_timeout(script)
+    wall_time = script.wall_time.to_i
+    return site_timeout if wall_time == 0
+    return [wall_time, site_timeout].min unless site_timeout == 0
+
+    wall_time
+  end
```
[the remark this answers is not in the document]
plazonic: I don't know about this one - it seems to me that the form.yml should not allow anything less than 1? I guess I'll take a look at how other adapters are doing it.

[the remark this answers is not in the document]
johrstrom: Yea mostly I think it's a scheduler thing. If you don't give a walltime there's some default. I checked linux_host and `timeout 0` must mean it's forever?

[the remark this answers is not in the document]
plazonic: I think you are correct - I changed it so that if it ends up being 0 then it will set the limit to infinity - which is what systemd wants in that case. Thanks!

lib/ood_core/job/adapters/linux_systemd/launcher.rb
```diff
@@ -163,10 +163,15 @@ def export_env(script)
...
   def script_timeout(script)
     wall_time = script.wall_time.to_i
-    return site_timeout if wall_time == 0
-    return [wall_time, site_timeout].min unless site_timeout == 0
-
-    wall_time
+    if wall_time == 0
+      # this is the only way it can be 0
+      # so make it into infinify for systemd to never terminate
+      site_timeout == 0 ? 'infinity' : site_timeout
```
johrstrom, line 169: ohhhh. @treydock do we want to allow infinity?  I'm kinda thinking yes even though sites may not expect it.

At least as unit files admins can find and terminate them easily, so they're discoverable (and not just through ps + grep).

johrstrom: Well actually we already allow it in the `linux_host` adapter (though I don't know if we knew that at the time).

It's just once we enable it we can't disable it without some pain for folks, so I want to be sure we're purposeful about it.

plazonic: I liked you pointing it out because I think some sites will want that - run until machine gets rebooted. Overall it is unlikely to be a problem because the default for min on bc_num_hours attribute is 1 - you have to go out of your way to allow 0. That is - as long as it is documented.

As far as how discoverable it is - you'd be surprised. I find it pretty hard to discover user units as root (by using systemctl).

plazonic:
Renamed. Thanks!

plazonic:
Please go ahead, I'm sure you can do it much much faster than I can. Thanks!

johrstrom:
Sorry I didn't see this before - but you're commits are not tied to your github user. I think it's an issue with the email + username combination?  I want to be sure you get credit.

plazonic:
If you mean that I didn't have that particular email address associated with my github account - I just added it. Does that make it look better? Thanks for thinking about it...

johrstrom:
That's it exactly. TYSM for the addition!

PR #748, opened 2022-03-01
Improved pod security

treydock:
Using Kyverno 1.6.0 pod security standards that mimic Kubernetes pod security standards.  We do have to filter out a few that simply can't support, like we do have to allow hostPath, but these are simple security changes to be able to conform to these new policies.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201901746971618) by [Unito](https://www.unito.io)

### DIFF
--- a/lib/ood_core/job/adapters/kubernetes/templates/pod.yml.erb
+++ b/lib/ood_core/job/adapters/kubernetes/templates/pod.yml.erb
@@ -106,7 +106,9 @@ spec:
       allowPrivilegeEscalation: false
       capabilities:
         drop:
-        - all
+        - ALL
+      seccompProfile:
+        type: RuntimeDefault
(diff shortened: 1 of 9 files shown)


johrstrom:
Are there any backward compatibility that you know off of hand?

treydock:
I believe the seccompProfile only works on Kubernetes 1.19 and newer, at least that's when it became stable in docs.  https://kubernetes.io/docs/tutorials/security/seccomp/

johrstrom:
Cool. 1.19 is already EOL so we're good.

treydock:
@johrstrom Could we get this into 2.0.x release so I can remove some temporary exclusions in our policies?

johrstrom:
Sure thing.

PR #752, opened 2022-03-24
Added cluster_info

lukew3:
Can't test current version because it uses the `call` method which requires loading the slurm config at `/etc/slurm/slurm.conf`, but was tested and works with regular backtick calls. I also would like to get feedback about my comments.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202021625654353) by [Unito](https://www.unito.io)


lib/ood_core/job/adapters/slurm.rb
```diff
@@ -98,6 +98,24 @@ def initialize(cluster: nil, bin: nil, conf: nil, bin_overrides: {}, submit_host
...
+            gpu_nodes_free = sinfo_out2.lines.uniq.grep(/gpu:/).grep(/idle/).count
+            {
+              "nodes_active": sinfo_out[0].to_i,
+              "nodes_total": sinfo_out[2].to_i,
+              "processors_total": sinfo_out[6].to_i,
+              "processors_active": sinfo_out[3].to_i,
+              "gpu_nodes_total": gpu_nodes_total,
+              "gpu_nodes_active": gpu_nodes_total - gpu_nodes_free
+            }
+          end
```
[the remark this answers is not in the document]
lukew3: > I've got to look into those sinfo calls a bit more to be sure that's what we're looking for.

The calls I implemented are the same as those found in the osc/osc-systemstatus repo.
Nodes/Processors - https://github.com/OSC/osc-systemstatus/blob/master/lib/slurm_squeue_client.rb#L175
Gpu - https://github.com/OSC/osc-systemstatus/blob/master/lib/slurm_squeue_client.rb#L111

lukew3:
Tested call methods, and are now working as intended

lib/ood_core/job/cluster_info.rb
```diff
@@ -0,0 +1,31 @@
...
+      def initialize(cluster_name: nil, nodes_active: nil, nodes_total: nil, processors_active: nil, processors_total: nil,
+                     gpu_nodes_active: nil, gpu_nodes_total: nil, **_)
+        @cluster_name        = cluster
+        @nodes_active        = nodes_active
+        @nodes_total         = nodes_total
+        @processors_active   = processors_active
+        @processors_total    = processors_total
+        @gpu_nodes_active    = gpu_nodes_active
+        @gpu_nodes_total     = gpu_nodes_total
+      end
```
johrstrom, line 17: I didn't want to type all these out, so here's just the one that you can repeat. (I also reversed the adjective + noun in the name).

Also  - you don't need cluster (or `cluster_name`). If I'm doing `clusters[:owens].job_adapter.cluster_info` I already know that it's the `owens` cluster.

```suggestion
      def initialize(opts = {})
        opts = opts.symbolize_keys
        @active_nodes        = opts.fetch(:active_nodes, nil).to_i
      end
```

johrstrom, line 17: I didn't want to type all these out, so here's just the one that you can repeat. (I also reversed the adjective + noun in the name).

Also  - you don't need cluster (or `cluster_name`). If I'm doing `clusters[:owens].job_adapter.cluster_info` I already know that it's the `owens` cluster.

```suggestion
      def initialize(opts = {})
        opts = opts.symbolize_keys
        @active_nodes        = opts.fetch(:active_nodes, nil).to_i
      end
```

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -98,6 +98,23 @@ def initialize(cluster: nil, bin: nil, conf: nil, bin_overrides: {}, submit_host
             @strict_host_checking = strict_host_checking
           end
 
+          # Get a ClusterInfo object containing information about the given cluster
+          # @return [ClusterInfo] object containing cluster details
+          def get_cluster_info
+            sinfo_out = call("sinfo", "-aho %A/%D/%C").strip.split('/')
+            sinfo_out2 = call("sinfo", "-Nhao %n/%G/%T").lines.uniq
```
johrstrom, line 105: these variables need better names. like `node_info` and `gres_info`?

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -98,6 +98,23 @@ def initialize(cluster: nil, bin: nil, conf: nil, bin_overrides: {}, submit_host
...
+            gpu_total = sinfo_out2.grep(/gpu:/).count
+            gpu_free = sinfo_out2.grep(/gpu:/).grep(/idle/).count
```
johrstrom, line 107: `sinfo -Nhao %n/%G/%T` I think this is just node status. I don't know if `allocated` means the GRES was also allocated. Meaning, I have a job on a node that has a GPU, but I didn't request or allocate it.


I think we actually want something more like `sinfo -Nha -O 'GresUsed'` that indicates whether the GRES is in use or not (as a pose to the the Node being in use or not).

  @treydock please advise -

johrstrom: Oh i see, we're returning `gpu_nodes_active`. No I think the better information is GPUs active. There are multiple GPUs on a given node.

I think reporting # of GPU Nodes is secondary to reporting on # of GPUs themselves.

treydock: Collecting GRES used with sinfo is actually a complicated process.  I ran into this with Slurm exporter.  The information about available/used GRES is only available via `--Format` flag so the printf `%G` isn't possible to get used, that's just available.

First you must get longest GRES line (they can get very long). Also you can likely cache this as GRES definitions only change when slurm.conf and gres.conf are changed and slurmctld is restarted.

```
sinfo -o '%G'
```

Then get GRES data:

```
sinfo -a -h --Node --Format=nodehost,gres:%d,gresused%d
```

You replace `%d` with value of longest GRES plus a few if you want a buffer.

treydock: In case not clear, the `sinfo -o '%G'` you have to iterate over all lines to find longest and use that for the length with `--Format`

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -98,6 +98,25 @@ def initialize(cluster: nil, bin: nil, conf: nil, bin_overrides: {}, submit_host
             @strict_host_checking = strict_host_checking
           end
 
+          # Get a ClusterInfo object containing information about the given cluster
+          # @return [ClusterInfo] object containing cluster details
+          def get_cluster_info
+            def gpus_from_gres(gres)
+              gres.to_s.scan(/gpu:[^,]*(\d+)/).flatten.map(&:to_i).sum
+            end
```
johrstrom, line 106: I don't think you need this redefinition here anymore.

lukew3: Regex still works on these cases #754. The regex gets the last set of digits before a comma or the end of rthe string if preceded by gpu:, so this holds.

lukew3: > I don't think you need this redefinition here anymore.

gpus_from_gres isn't defined here since it's defined in Slurm while get_cluster_info is defined in Batch, which is nested in Slurm. Should Batch extend Slurm?

johrstrom: I see, we can delegate the method to the member variable (`@slurm`) so that it looks as if it's coming from the `Slurm` class.

```ruby
      class Slurm < Adapter
        using Refinements::HashExtensions
        using Refinements::ArrayExtensions

        delegate :gpus_from_gres  to: :slurm
```

It's ruby's synactic sugar for

```ruby 
def gpus_from_gres 
   @slurm.gpus_from_gres 
end
```

lukew3: > I see, we can delegate the method to the member variable (`@slurm`) so that it looks as if it's coming from the `Slurm` class.

The `delegate` method is included in rails, but not ruby. It looks like `def_delegators` from ruby's `forwardable` module should be able to do the same thing, but I can't get that to work either.

johrstrom: Lol that's hilarious. This'll do just fine in a pinch.

```ruby
def gpus_from_gres 
   @slurm.gpus_from_gres 
end
```

lukew3: I actually don't think that delegate is right for this. From [rubydocs](https://www.rubydoc.info/docs/rails/4.1.7/Module:delegate) `delegate`:
> Provides a delegate class method to easily expose contained objects' public methods as your own.

So I would want this if I wanted to call a Batch method from Slurm, when I actually need to call a Slurm method from Batch.

lukew3: > This'll do just fine in a pinch.
> 
> ```ruby
> def gpus_from_gres 
>    @slurm.gpus_from_gres 
> end
> ```

That doesn't work either:
```
undefined method `gpus_from_gres' for nil:NilClass
```

johrstrom: Right because I've got it backwards. Yea I'd say let's move that function into `Batch` then just reference it through `@slurm.gpus_from_gres`. Other adapters have a helper class which is where it'd go, but alas, we did not use that here.

treydock: > The regex gets the last set of digits before a comma or the end of the string if preceded by gpu:, so this holds.

But not sure that will work, because the gres won't always be proceeded by `gpu:<digit>`, it might be `gpu:a100:<digit>` because a GRES can have a subtype and users can request sub-types so that would be what shows up.

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -454,6 +473,12 @@ def submit(script, after: [], afterok: [], afternotok: [], afterany: [])
           raise JobAdapterError, e.message
         end
 
+        # Retrieve info about active and total cpus, gpus, and nodes
+        # @return [Hash] information about cluster usage
+        def cluster_info()
+          @slurm.get_cluster_info()
+        end
```
johrstrom, line 480: The ruby-ish way to write this is without parentheses. 

```suggestion
        def cluster_info
          @slurm.get_cluster_info
        end
```

lib/ood_core/job/adapters/slurm.rb
lukew3: > Right because I've got it backwards. Yea I'd say let's move that function into `Batch` then just reference it through `@slurm.gpus_from_gres`. Other adapters have a helper class which is where it'd go, but alas, we did not use that here.

That doesn't work either. Slurm.parse_job_info can't load `@batch.gpus_from_gres` because it can't access batch. To fix this, I defined `gpus_from_gres` just above the Batch class definition, so that it is accessible everywhere in both classes.

PR #753, opened 2022-03-24
Added uses_gpu flag to Job Info class

lukew3:
Test cases we want to be sure to cover:
- [ ] gres is `nil`
- [ ] gres is "" (empty)
- [ ] gres is `gpu:v100-32g:2`
- [ ] gres is `gpu:v100-32g:2,pfsdir:1`
- [ ] gres is `third-thing:sub-thing:17,gpu:v100-32g:2,pfsdir:1`
- [ ] gres is `third-thing:sub-thing:17,pfsdir:1,gpu:v100-32g:2`
- [ ] gres is `gres:gpu:1`
- [ ] gres is `gres:pfsdir:ess`

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1202023086162926) by [Unito](https://www.unito.io)


lib/ood_core/job/info.rb
```diff
@@ -111,6 +115,7 @@ def initialize(id:, status:, allocated_nodes: [], submit_host: nil,
         @status = job_array_aggregate_status unless @tasks.empty?
 
         @native          = native
+        @uses_gpu        = uses_gpu || native && native[:gres] ? native[:gres].include?("gpu") : false
```
johrstrom: The calculation shouldn't be here because it will be adapter specific. You can add the `slurm` specifics in this PR - but basically when this object is built, the builder determines the flag and passes it in.

Also now -rethinking this, a job can request more than 1 gpu so - this should reflect that. I.e., the attribute is the number of gpus for the job ranging from 0 - N.

```ruby
   # the number of gpu's the job has requested
   attr_reader :gpus

   def initialize(
     # ...
     @gpus        =  gpus.to_i
   end
        
    def gpu?
       gpus.positive?
    end
```

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -643,7 +643,8 @@ def parse_job_info(v)
               cpu_time: nil,
               submission_time: v[:submit_time] ? Time.parse(v[:submit_time]) : nil,
               dispatch_time: (v[:start_time].nil? || v[:start_time] == "N/A") ? nil : Time.parse(v[:start_time]),
-              native: v
+              native: v,
+              gpus: v[:gres]&.include?("gpu") ? 1 : 0
```
johrstrom, line 647: We probably want some sort of regular expression here to extract a number to get the right number.

treydock: The format for GRES is `gres:subtype:number` so if you split on colons and get last element, that should be the number.  I would not recommend extracting digits because at OSC, we have numbers in GRES subtype like we have `gpu:v100:2` or `gpu:v100-32g:2`.  Maybe `([^:]+):([^:]+):([0-9]+)` if you really want to use regex.

treydock: Ah my regex is a little too simple, the `subtype` is optional so you can have `gres:number` or `gres:subtype:number`.  So I think in those cases just splitting on colons and pulling out the last element might be simplest to avoid overly complex regex.

johrstrom: We'll need at least the `gpu` to check against others like `pfsdir` and so on.  

Maybe `[,]*gpu.*:(\d+)[,]*`?  @lukew3 I'll update this ticket for test cases we want to be sure to cover.

lukew3: > Maybe `[,]*gpu.*:(\d+)[,]*`? @lukew3 I'll update this ticket for test cases we want to be sure to cover.

I split by commas and colons to count it, so this works:
```ruby
gres&.split(',')&.each { |s| gpus += s.split(':').last.to_i if s.split(':')[0] == 'gpu' }
```

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -617,6 +617,12 @@ def get_state(st)
             STATE_MAP.fetch(st, :undetermined)
           end
 
+          def get_gpu_count(gres)
+            gpus = 0
+            gres.to_s.scan(/gpu:[^,]*(\d+)/).each { |val| gpus += val[0].to_i }
+            gpus
```
johrstrom, line 623:
```suggestion
            gres.to_s.scan(/gpu:[^,]*(\d+)/).flatten.map(&:to_i).sum
```

Anytime you find yourself doing something like this - initializing outside a loop then returning that - in Ruby there's likely a much better way.

johrstrom: Specifically like `inject` or `each_with_object`, but in this case, we can directly just remap and sum the array.

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -617,6 +617,10 @@ def get_state(st)
             STATE_MAP.fetch(st, :undetermined)
           end
 
+          def get_gpu_count(gres)
+            gres.to_s.scan(/gpu:[^,]*(\d+)/).flatten.map(&:to_i).sum
+          end
```
johrstrom, line 622: Sorry! :sweat_smile: Ruby's parlance is very declarative, so you don't see a lot of `get` or `set` prefixes.

```suggestion
          def gpus_from_gres(gres)
            gres.to_s.scan(/gpu:[^,]*(\d+)/).flatten.map(&:to_i).sum
          end
```

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -643,7 +647,8 @@ def parse_job_info(v)
               cpu_time: nil,
               submission_time: v[:submit_time] ? Time.parse(v[:submit_time]) : nil,
               dispatch_time: (v[:start_time].nil? || v[:start_time] == "N/A") ? nil : Time.parse(v[:start_time]),
-              native: v
+              native: v,
+              gpus: get_gpu_count(v[:gres])
```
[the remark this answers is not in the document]
johrstrom: GRES is the very last column in this output (currently `(null)`). Can you update one of them with a real live example and update the test cases that use that fixture file to verify it's being set.

https://github.com/OSC/ood_core/blob/fd26e1622ab714bfa89fd758599445e60b292297/spec/fixtures/scripts/squeue.rb#L4-L5


Here's where you can verify it on the other side. Note that both j1 and j2 should make assertions on `gpus`  and `gpu?`.
https://github.com/OSC/ood_core/blob/7aaad325f16aac5e1b8fb11cbc4dfe95f36b4075/spec/job/adapters/slurm_spec.rb#L308

johrstrom (approved):
Thanks, I think we're there.

PR #762, opened 2022-06-07
Fix gpus_from_gres unreadable from Slurm::Batch job adapter

lukew3:
As of v0.20.0, `cluster_info` is unable to be retrieved as `gpus_from_gres` is inaccessible from Slurm::Batch. Making `gpus_from_gres` a class method of Slurm and calling `Slurm.gpus_from_gres` inside of Batch corrects this. Not sure how this issue came to be as I remember it working when we merged.


johrstrom (approved):
Works for me.

PR #767, opened 2022-07-19
check for nohttpd option before using it with vncserver

### LINKED ISSUE #737: TurboVNC 2.2.90 lacks the -nohttpd option
Using the TurboVNC repo and pulling the 'turbovnc' package results in a version of TurboVNC on Rocky 8 (and I assume other RHEL 8 variants) which fails when called with -nohttpd, breaking the default bc_desktop app, thanks to its inclusion in vnc.rb - http server support has been dropped for newer versions of TurboVNC, though it will be supported in 2.2.x versions. (See release notes here: https://github.com/TurboVNC/turbovnc/releases/tag/3.0beta1) 

Not sure what the most sustainable fix is, perhaps just mentioning this in the documentation at https://osc.github.io/ood-documentation/latest/app-development/interactive/setup/enable-reverse-proxy.html... This is really just a heads up - possible to continue with new installs using an older TurboVNC version just fine.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201625178070145/1201627216789264) by [Unito](https://www.unito.io)

johrstrom:
Fixes #737

### DIFF
--- a/lib/ood_core/batch_connect/templates/vnc.rb
+++ b/lib/ood_core/batch_connect/templates/vnc.rb
@@ -86,7 +86,7 @@ module OodCore
                 # Clean up any old VNC sessions that weren't cleaned before
                 #{vnc_clean}
 
-                #for turbovnc 3.0 compatability.
+                # for turbovnc 3.0 compatability.
                 if timeout 2 vncserver --help 2>&1 | grep 'nohttpd' >/dev/null 2>&1; then
                   HTTPD_OPT='-nohttpd'
                 fi


lib/ood_core/batch_connect/templates/vnc.rb
```diff
@@ -86,8 +86,13 @@ def before_script
                 # Clean up any old VNC sessions that weren't cleaned before
                 #{vnc_clean}
 
+                #for turbovnc 3.0 compatability.
+                if timeout 2 vncserver --help 2>&1 | grep 'nohttpd' >/dev/null 2>&1; then
```
treydock, line 90: Sure be nice if there was a way to get the version, but just looked and appears the module installed on OSC systems does not have any kind of version flag.

johrstrom: Lol yea that's why I went for this version. Also may be compatible with `tigervnc`? I've seen folks have that installed instead and it cause issues.

treydock: We install tigervnc on bastion servers:

```
$ vncserver --help

usage: vncserver [:<number>] [-name <desktop-name>] [-depth <depth>]
                 [-geometry <width>x<height>]
                 [-pixelformat rgbNNN|bgrNNN]
                 [-fp <font-path>]
                 [-cc <visual>]
                 [-fg]
                 [-autokill]
                 [-noxstartup]
                 [-xstartup <file>]
                 [-fallbacktofreeport]
                 <Xvnc-options>...

       vncserver -kill <X-display>

       vncserver -list
```

Looks like your logic would work there too.

johrstrom:
I've got a ticket with Sci apps to get 3.0 on OSC systems. Would like to test this first, but I'd also like to get it into the next 2.0 release because it's a real pain point for folks.

treydock (approved):

johrstrom:
I was able to confirm this works with a module I built myself. The script_wrapper had `set -x` so here's the output of the 2 cases:

With 2.2.6
```
+ grep nohttpd
+ timeout 2 vncserver --help
+ HTTPD_OPT=-nohttpd
++ vncserver -log vnc.log -rfbauth vnc.passwd -nohttpd -noxstartup -geometry 1536x864
+ VNC_OUT='
```

with 3.0
```
+ vncserver -list
+ grep nohttpd
+ timeout 2 vncserver --help
++ vncserver -log vnc.log -rfbauth vnc.passwd -noxstartup -geometry 1536x864
+ VNC_OUT='
```

PR #778, opened 2022-09-27
add k8s ability to read many items from secrets

### LINKED ISSUE #744: kubernetes support conn_params
Similar to https://github.com/OSC/ondemand/issues/1829 connection info here needs to support arbitrary things added to it to support `conn_params`. I think keeping things in a secret here is our best bet (even if they aren't secret).

We're extracting password as a default - but it seems like we could inspect this and extract more.

https://github.com/OSC/ood_core/blob/b4b21298936a744c0d552e5e52acbafaeb57bb07/lib/ood_core/job/adapters/kubernetes/helper.rb#L195-L200

 Though I don't know if we'll have to pass back everything because I'm not sure we'll know what `conn_params` are defined. (maybe in the submit we set them to empty strings with known names like `conn_params/csrf_token` if I wanted to pass back `csrf_token`).

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1201735133575781/1201794991434315) by [Unito](https://www.unito.io)

johrstrom:
Fixes #744 such that k8s jobs can add as many secrets as they like and we'll read them all back as key value pairs.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203058034806804) by [Unito](https://www.unito.io)

### DIFF
--- a/lib/ood_core/job/adapters/kubernetes/helper.rb
+++ b/lib/ood_core/job/adapters/kubernetes/helper.rb
@@ -195,10 +195,14 @@ class OodCore::Job::Adapters::Kubernetes::Helper
   end
 
   def secret_info_from_json(json_data)
-    raw = json_data.dig(:data, :password)
-    { ood_connection_info: { password: Base64.decode64(raw) } }
-  rescue
-    {}
+    data = json_data.to_h[:data] || {}
+
+    info = data.symbolize_keys.each_with_object({}) do |data_kv, hash|
+      hash[data_kv[0]] = Base64.decode64(data_kv[1])
+    rescue
+      next
+    end
+    { ood_connection_info: info }
   end
 
   def dispatch_time(json_data)


lib/ood_core/job/adapters/kubernetes/helper.rb
```diff
@@ -195,10 +195,14 @@ def service_info_from_json(json_data)
...
-  rescue
-    {}
+    data = json_data.to_h[:data] || {}
+
+    info = data.symbolize_keys.each_with_object({}) do |data_kv, hash|
+      hash[data_kv[0]] = Base64.decode64(data_kv[1])
+    rescue
+      next
+    end
+    { ood_connection_info: info }
```
Oglopf, line 205: When this returns the `info` will be a hash itself with many password keys and values. Will that be an issue since previously we returned a hash with a single `password` key? 

It looks good, just trying to think if that would cause an issue of a returning hash now having many `password` keys that previously had been used with just the one `password` key.

Oglopf: Ok, after looking at the code closer I see this won't be an issue. The method is `private` and being used later with a `deep_merge` into the `pod_hash` which handles all this fine.

Looks good!

Oglopf (approved):
LGTM.

PR #783, opened 2022-12-05
get all available accounts for a user from the scheduler

johrstrom:
Fixes upstream ood issue below. The idea here is that the scheduler (Slurm in this case) knows all the valid accounts that a user is able to use. So instead of pulling accounts from Unix file systems, we can query the scheduler itself. A Unix group may or may not be a valid account as the account can suspended/closed due to budgets but the Unix group still exists.

https://github.com/OSC/ondemand/issues/1970

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203495593751580) by [Unito](https://www.unito.io)


lib/ood_core/job/adapters/slurm.rb
```diff
@@ -178,6 +179,18 @@ def get_jobs(id: "", owner: nil, attrs: nil)
             return [{ id: id, state: 'undetermined' }]
           end
 
+          def accounts
+            user = Etc.getlogin
+            args = ['-np', 'show', 'accounts', 'withassoc', 'format=account,user', '-P']
```
treydock, line 184: This will take a VERY long time on OSC as we have thousands of accounts and even more users.  You will almost certainly want the sacctmgr command to have `where user=<username>` as an argument otherwise you dump all users which at OSC could take a minute or two or even longer.

treydock: Also why use `-p` and `-P` in same command, it's either one of the other , not both.  Using just `-p` has trailing pipe where `-P` does not.

Ran this command at OSC and faster than expected but still took 5 seconds.  But using my username of `where user=tdockendorf` took 0.5 seconds.

treydock (changes requested):
See inline comments for requested changes

lib/ood_core/job/adapters/slurm.rb
treydock, line 184:
```suggestion
            args = ['-nP', 'show', 'users', 'withassoc', 'format=account,user', 'where', "user=#{user}"]
```

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -178,6 +179,18 @@ def get_jobs(id: "", owner: nil, attrs: nil)
...
+
+            [].tap do |accts|
+              call('sacctmgr', *args).each_line do |line|
+                acct, username = line.split('|')
+                accts << acct if username.to_s.chomp == user
+              end
+            end.uniq.map(&:upcase)
```
treydock, line 191: Is it possible to not do upcase but only do upcase for OSC? I ask because Slurm down cases everything in the database, both users and accounts.

johrstrom: Yea I was also wondering how to do this. some sort of environment variable `OOD_UPCASE_SLURM_ACCOUNTS` or similar?

lib/ood_core/job/adapters/slurm.rb
johrstrom: Thanks, I'm going to commit the suggestion and then fix the tests it likely breaks.

johrstrom: Yea thanks for those both, especially the `where` clause as I was seeing it take some time (was going to cache it in the OOD layer).

lib/ood_core/job/adapters/slurm.rb
treydock: Ya an environment variable or cluster YAML config option.

johrstrom: Went for an environment variable as that's an easier interface (this library doesn't read any config files yet safe for cluster files).

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -178,6 +179,21 @@ def get_jobs(id: "", owner: nil, attrs: nil)
...
+            args = ['-nP', 'show', 'users', 'withassoc', 'format=account,user', 'where', "user=#{user}"]
+            upcase = ENV['OOD_UPCASE_ACCOUNTS'].nil? ? false : true
+
+            [].tap do |accts|
+              call('sacctmgr', *args).each_line do |line|
+                acct, _ = line.split('|')
+                accts << acct unless acct.nil?
+              end
+            end.uniq.map do |acct|
+              upcase ? acct.upcase : acct
```
Oglopf, line 193:
```suggestion
              :upcase ? acct.upcase : acct
```
I think this needs to be a symbol, wouldn't work for me locally without this.

johrstrom: That's odd... it should be a boolean.

johrstrom: I don't follow. Did you get an error or did the tests just fail?

```
irb(main):007:0> f = false
=> false
irb(main):008:0> :f ? true : false
=> true
irb(main):009:0> f ? true : false
=> false
```

Oglopf: I just mocked the array of strings and ran it through using the pattern you had but noticed I got a failure without the symbol is all. I guess I would always think in Ruby to do this you'd need to use that symbol pattern:
```
['one', 'two', 'THREE'].map do |str|
  :upcase ? str.upcase : str
end
```
Was what I did to mock, but I may be misunderstanding this code.

Oglopf: Ok, I see now there's an `upcase` variable doing something with a ternary above this. I need to take this apart more to understand all that's going on.

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -483,6 +499,13 @@ def cluster_info
           @slurm.get_cluster_info
         end
 
+        # Retrieve the accounts available to use  for the current user.
+        #
+        # @return [Array<String>] the accounts available to the user.
+        def accounts
+          @slurm.accounts
+        end
```
Oglopf, line 507: Why is there a `private and `public` version of this method?

johrstrom: I can't tell the scope from here, but there's the outer class `Slurm < JobAdapter` and an inner class. What you've highlighted is the outer most public class. But then there's the inner object we call `@slurm` which does all the actual work.

You can see the `Batch` class (the type of object that `@slurm` is) also defined here.

Oglopf: Oh, yeah, this file is hideous. Yeah sorry, i saw the private and just thought it was in that scope....but that ends, then we have an `initialize` then this.....ok, I really can't understand this design tbh so I need to talk through this.

The class instantiates itself then uses that object to call its methods on itself to build this? Sorry, I am a bit lost in this adapter.

johrstrom: No it's just a nested class. I.e., a class defined within the scope of another.

Here's a simple example where you'd interact with `Outer` objects without caring about the `Inner` or any `Inner` instantiations. 
```ruby
class Outer
  def initialize
    @inner = Inner.new()
  end

  def do_something
    @inner.do_something
  end

  class Inner
    def do_something
      puts 'here I am'
    end
  end
end

Outer.new.do_something
```

Why not just put all the logic in `Outer`? The pattern was started before my time, but it's common job adapter pattern where we have another `Batch` class that actually does all the work.

It's just with the other adapters, the `Batch` classes are in other files.  They have the same pattern, here's LSF as an example where it's using an inner object called `batch`.

https://github.com/OSC/ood_core/blob/dc7bfbdc622f851194e7c7b56dadb3a61f466c95/lib/ood_core/job/adapters/lsf.rb#L106-L110

Oglopf: Ah, ok, I've seen that pattern, I'm just lost in indentation in this file. Ok, that makes sense and gets this cleared up.

Oglopf (approved):
Looks good.

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -178,6 +179,21 @@ def get_jobs(id: "", owner: nil, attrs: nil)
             return [{ id: id, state: 'undetermined' }]
           end
 
+          def accounts
+            user = Etc.getlogin
+            args = ['-nP', 'show', 'users', 'withassoc', 'format=account,user', 'where', "user=#{user}"]
+            upcase = ENV['OOD_UPCASE_ACCOUNTS'].nil? ? false : true
```
treydock, line 185: I doubt this will happen but if someone says `OOD_UPCASE_ACCOUNTS=false` this would evaluate to `true`. Not sure if that's desirable.

johrstrom:
I don't believe this will work as is. We're likely going to need at least `qos` associated with the account. So passing just strings back is not likely enough.

With some sites, `partition` will show up in this output. On other sites, I believe we'll need another query to stitch together with this data.
```
sacctmgr show association 'format=account,partition,qos%100,' where "user=$(whoami)"
```

treydock:
If you use `--parsable` or `--parsable2` flags you don't need format length like `%100`. Just FYI as that would simplify the logic.  

As far as whether this works, I think it's a good starting place.  Having a list of available accounts irregardless of partition is good. However I don't think you will need to query QOS since an association in the database is either "cluster + account + user" or "cluster + account + user + partition".  The QOS is an attribute on the association.  I think it's a bridge too far for OOD to try and work out how a QOS is used because a QOS can be used to limit resources or just do things like set priority or preemption.  A QOS is not typically how you grant someone access to resources, that's usually done through the association in one of the two defining ways previously mentioned.

treydock (approved):

johrstrom:
Thanks for the reivew, I've added a bit. Namely the AccountInfo class so it can hold extra information for all the use cases we need.

Check out this Utah sacctmgr output. They use 1 account per cluster and that account only exists on that cluster.  So for this to be useful every account has to be sort of cluster aware so that upper layers can hide/show options as needed.
```
[u6031686@kingspeak1:~]$ sacctmgr -np show user withassoc format=account,qos,cluster where user=$(whoami)
smithp-guest|ash-guest,ash-guest-res|ash|
chpcservice|kingspeak|kingspeak|
owner-guest|kingspeak-guest|kingspeak|
chpcservice|lonepeak|lonepeak|
owner-guest|lonepeak-guest|lonepeak|
dtn|notchpeak-dtn|notchpeak|
notchpeak-shared-short|notchpeak-shared-short|notchpeak|
chpcservice|notchpeak-freecycle|notchpeak|
owner-guest|notchpeak-guest|notchpeak|
```

I'm quite sure this is not enough as the upper layers (OOD) is going to have to stitch a lot of information together. But as an initial luanch maybe it can just pull `adapter.accounts.map(&:to_s)` initially. 

In any case, I know for sure these are going to be more complex than just a string, so we may as well start with objects.

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -325,7 +342,7 @@ def call(cmd, *args, env: {}, stdin: "")
               cmd = OodCore::Job::Adapters::Helper.bin_path(cmd, bin, bin_overrides)
 
               args  = args.map(&:to_s)
-              args.concat ["-M", cluster] if cluster
+              args.concat ["-M", cluster] if cluster && cmd != 'sacctmgr'
```
johrstrom, line 345: Unfortunately I can't seem to get `sacctmgr` to query per cluster, which is going to be troublesome for upper layers as each adapter object is meant to represent 1 cluster, but here we are returning information for all clusters.

PR #795, opened 2023-02-07
ACLs to use block & allowlists

johrstrom:
ACLs to use block & allowlists instead of white & black lists. This is still backward compatible and should be for some time after we update the 2.1 documentation.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/1135148780858012/1203923040787317) by [Unito](https://www.unito.io)


gerald-byrket (approved):
This looks good.  I see you are testing for all 4 groups of wording for backwards compatibility.  Thanks for doing that.

PR #811, opened 2023-07-31
find_port: avoid infinite loop

utkarshayachit:
If all ports in chosen port range were taken, `find_port` had the tendency to block. This fix updates the code to avoid this and instead attempt to find a port by looping over the port range for a fixed number of retries.


johrstrom (approved):
Thanks for the bug fix! Unfortunately we don't have strong test cases for this, but I was able to pull the modification into a simple bash script and test it and it's good.

PR #813, opened 2023-07-31
vnc: run websockify as background process

utkarshayachit:
Instead of running as a daemon, making websockify run as a background process. This ensures that the process gets cleaned up property if and when job is terminated. With certain job schedulers/environments (e.g. OpenPBS+AzHOP), this avoids dangling processes lingering post job termination.

### DIFF
--- a/lib/ood_core/batch_connect/templates/vnc.rb
+++ b/lib/ood_core/batch_connect/templates/vnc.rb
@@ -134,11 +134,45 @@ module OodCore
             <<-EOT.gsub(/^ {14}/, "")
               #{super}
 
+              # launches websockify in the background; waiting until the process
+              # has started proxying successfully.
+              start_websockify() {
+                local log_file="./websockify.log"
+                # launch websockify in background and redirect all output to a file.
+                #{websockify_cmd} $1 $2 &> $log_file &
+                local ws_pid=$!
+                local counter=0
+
+                # wait till websockify has successfully started
+                echo "[websockify]: pid: $ws_pid (proxying $1 ==> $2)" >&2
+                echo "[websockify]: log file: $log_file" >&2
+                echo "[websockify]: waiting ..." >&2
+                until grep -q -i "proxying from :$1" $log_file
+                do
+                  if ! ps $ws_pid > /dev/null; then
+                    echo "[websockify]: failed to launch!" >&2
+                    return 1
+                  elif [ $counter -ge 5 ]; then
+                    # timeout after ~5 seconds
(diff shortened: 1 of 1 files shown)


utkarshayachit:
not doing the same of `vnc_container` since there the issue of potential dangling processes won't exist since when the container is terminated, so will the `websockify` process.

johrstrom:
OK - I'll have to pull this down and take a closer look at it. BTW - it looks like it conflicts with the other PR I just merged.

utkarshayachit:
sure thing...I've pushed an updated to resolve the conflict as well. BTW, in my experiments I was added a cleanup code in `cleanup_sscript` to find and the started websockify process as well. While not strictly necessary, perhaps makes sense to add that to cleanup everything gracefully, what do you think.

here's the snipped. If you think it is reasonable I can update the MR to add a cleaned up version of the same:

```bash
find_websockify_pid() {
  local ws_port=$1
  # this matches command line used in start_websockify().
  local command_line="#{websockify_cmd} $ws_port"
  local pid=$(ps -eo pid,cmd | grep "$command_line" | grep -v grep | awk '{print \$1}')
  echo "$pid"
}
# cleanup websockify, if started
if [ -n "$websocket" ]; then
  ws_pid=$(find_websockify_pid ${websocket})
  if [ -z "$ws_pid" ]; then
    echo "[warning] websockify process not found!" >&2
  else
    echo "[info] killing websockify [pid=$ws_pid]" >&2
    kill $ws_pid
  fi
fi
```

johrstrom:
> While not strictly necessary, perhaps makes sense to add that to cleanup everything gracefully, what do you think.

Errmm... I'm not sure. Or at least I don't like grepping through ps output to find it.  If we do, or need to, maybe we should save the PID to a `$PWD/.websockify_pid` or similar (quickly tried to look through `/proc/self`, but couldn't find anything).

utkarshayachit:
> Errmm... I'm not sure. Or at least I don't like grepping through ps output to find it.

Roger that. We can table that for now. I'll try to investigate in due time to see if there are better ways of getting the websockify PID.

utkarshayachit:
the latest revision looks at `websockify` output to ensure the process is started successfully as well.

johrstrom (approved):
Hi, sorry for the delay. With the new semester starting, I had a lot of OSC work to complete.

In any case, I've looked this over now and it appears to work well. Thanks for the addition!

utkarshayachit:
no problem at all! Thanks for the review/merge.

PR #815, opened 2023-11-02
Add working_dir option for Fujitsu TCS job scheduler

mnakao:
Dear developers,

I add working_dir option for Fujitsu TCS job scheduler, which is used in Fujitsu servers (the supercomputer Fugaku, Wisteria/BDEC-01 in the university of Tokyo, and so on).

The directory where the adapter is run is `/var/www/ood/apps/sys/dashboard`, but we found that running it in this location may cause an error on some Fujitsu systems. Therefore, we added an option for working_dir, allowing you to change the directory where the adapter runs.

This is the same as slurm's `sbatch -D` option, but since there is no equivalent to the `-D` option in Fujitsu TCS, the working_dir option is set in the cluster definition file (e.g. `/etc/ood/config/clusters.d/fugaku.yml`).

Best,


mnakao:
Sorry. I miss the file. I want to cancel this PR.

PR #820, opened 2023-11-28
Revert "Changes default return value for cluster.batch_connect_ssh_al…

HazelGrant:
…low? (#818)"

This reverts commit 6504b5f3fc175d9dec496905c4b8ce945db49149.


johrstrom:
Thanks for getting this ready for us even if we may not need it.  I'm still trying to reason about what the most appropriate precedence should be.

johrstrom:
I think we will be moving forward with this - can you update the changelog and version while we're at it so we can just release afterwards?

johrstrom (approved):

PR #826, opened 2024-02-29
Add configurable websocket heartbeat

scratchings:
Without the --heartbeat option the VNC session will timeout after 1 minute of idle time. This adds --heartbeat with a default of 30 seconds, configurable with websockify_hb in the context or the environment variable WEBSOCKIFY_HB.

This implements the VNC timeout workaround mentioned in 
[VNC timeout](https://discourse.openondemand.org/t/novnc-and-shell-session-timeout-after-1-minute/1622)

The VNC app in OSC/ondemand will need to be updated to recognise the websockify_hb context variable.

### DIFF
--- a/lib/ood_core/batch_connect/templates/vnc.rb
+++ b/lib/ood_core/batch_connect/templates/vnc.rb
@@ -130,7 +130,8 @@ module OodCore
           # successful connections so that the password can be reset
           def after_script
             websockify_cmd = context.fetch(:websockify_cmd, "${WEBSOCKIFY_CMD:-/opt/websockify/run}").to_s
-
+            websockify_hb = context.fetch(:websockify_hb, "${WEBSOCKIFY_HB:-30}").to_s
+            
             <<-EOT.gsub(/^ {14}/, "")
               #{super}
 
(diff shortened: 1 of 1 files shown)


johrstrom:
First off, yes please and thank you! I was going to patch this for the next release, so thank you for beating me to it!

lib/ood_core/batch_connect/templates/vnc.rb
```diff
@@ -130,7 +130,8 @@ def run_script
           # successful connections so that the password can be reset
           def after_script
             websockify_cmd = context.fetch(:websockify_cmd, "${WEBSOCKIFY_CMD:-/opt/websockify/run}").to_s
-
+            websockify_hb = context.fetch(:websockify_hb, "${WEBSOCKIFY_HB:-30}").to_s
```
johrstrom, line 133: I'm wondering if this should be expanded to `websockify_heartbeat_seconds` just to be a little more specific?  The internal variable doesn't matter as much to me, but the user facing configuration I think may need to be a little more explicit?

scratchings: Fine by me. I did consider this, but left it similar to the previous line where command has been abbreviated to 'cmd'

johrstrom: I was considering the same lol. `cmd` was before my time, but to an admin configuring it - it may be more recognizable then `hb`.

johrstrom: I'm going to go ahead and merge this now and change it later. Thanks again for the contribution!

johrstrom (approved):

PR #834, opened 2024-03-29
systemd adapter fails with zsh as user shell

m0zes:
zsh interprets the * in ondemand-* as a file path and returns an error if nothing is matched (unless the user has set 'setopt no_nomatch' in their .zshrc. Escape it to pass the asterisk on to systemctl (so it can be parsed internally).


johrstrom:
Thanks for the fix! I'll have to test this out a bit.

johrstrom:
Sorry for the delay - honestly i just plain forgot this was out there. I'll be sure to look at it next week.

johrstrom (approved):
Thanks for this bugfix. I apologize for the long delay for me to check it out.

PR #841, opened 2024-06-26
add rake task for getting squeue text output

johrstrom:
I went to write a test case for squeue in slurm and needed some squeue output to test against. Given that we expect the output to be of a certain format, this was very hard to do by hand.

So I've written this rake task to output 1 single job from squeue -  in the format that this library expects - to write to a fixture file that we can then test against.


Oglopf (approved):
Works for me.

PR #842, opened 2024-07-02
force utf-8 encoding

johrstrom:
force utf-8 encoding for Slurm to prevent errors when squeue returns ASCII characters that don't parse correctly.


spec/job/adapters/slurm_spec.rb
```diff
@@ -770,6 +770,29 @@ def job_info(opts = {})
...
+        "123"
+      ]}
+
+      it "correctly handles non utf8 characters" do
+        stdout = File.read('spec/fixtures/output/slurm/non_utf8_job_name.txt')
+        stdout.force_encoding(Encoding::ASCII)
+        allow(Open3).to receive(:capture3).with({}, *squeue_args, stdin_data: "").and_return([stdout, '', double("success?" => true)])
+        job = OodCore::Job::Factory.build_slurm({}).info('123')
+        expect(job.job_owner).to eq('annie.oakley')
+        expect(job.job_name).to eq('��� non-utf8')
```
johrstrom, line 793: There's some issue with stubbing the output data here where we expect 3 of these characters instead of one. In any case, comment out the line `line = line.encode('UTF-8', invalid: :replace, undef: :replace)` in the slurm adapter and it does indeed fail with the same `ArgumentError: invalid byte sequence in US-ASCII` so the test case does indeed cover that.

ashton22305 (approved):
LGTM

PR #850, opened 2024-08-08
add tests for ruby 3.1 and 3.2

johrstrom:
add tests for ruby 3.1 and 3.2

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,6 +18,10 @@ jobs:
         include:
           - ruby: "3.0.2"
             bundler: "2.3.5"
+          - ruby: "3.1"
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -18,6 +18,10 @@ jobs:
         include:
           - ruby: "3.0.2"
             bundler: "2.3.5"
+          - ruby: "3.1"
```
Oglopf, line 21: Do we want these pegged to a minor version as well? I'm fine being flexible and running against the latest minor version as I doubt there would be a breaking change, but I did want to make sure we are ok with that.

Oglopf (approved):
Looks good.

PR #851, opened 2024-08-09
add support for specifying cores for the adapters that can support it

johrstrom:
Add support for specifying cores for the adapters that can support it.

Fujitsu and SGE don't have support for this yet.  SGE seems to need a variable to specify the parallel environment, which is unknown upfront. We'd have to issue a `qconf` command or similar to find the applicable string to use.  Fujitsu I just can't seem to find the documentation for it.

I'll keep this open for just a second to write some tests for it.


mnakao:
@johrstrom 
Looking at the changes to slurm, I saw that "-n" had been added, so I thought that the number of tasks (number of MPI processes) would be specified.
If so, use the syntax --mpi proc=X for Fujitsu TCS.

`args.concat ["--mpi", "proc=" + script.cores] unless script.cores.nil?`
 
For details, please refer to the Section 2.3.5.2 in "Job Operation Software End-user's Guide" at the URL below.
https://software.fujitsu.com/jp/manual/manualindex/p21000155e.html

Thanks,

johrstrom:
Thanks! I'll update the fujitsu adapter and keep that documentation handy for other updates.

HazelGrant (approved):

Issue #852, opened 2024-08-15
Kubernetes Per User Resource does not support "." in user names. 


yhal-nesi:
Dot is a perfectly legal and common character in posix username, but it is used to generate a kubernetes namespace name and kubernetes doesn't allow dots.

Looks like there is one place where it needs to be adjusted?
https://github.com/OSC/ood_core/blob/master/lib/ood_core/job/adapters/kubernetes/batch.rb#L274

as well as the bootstrap script https://github.com/OSC/ondemand/blob/master/hooks/k8s-bootstrap/k8s-bootstrap-ondemand.sh

johrstrom:
hi thanks for the bug report. Seems like we need to sanitize to `-`?

yhal-nesi:
An example of how JupyterHub deals with K8 name generation: https://github.com/jupyterhub/kubespawner/blob/main/kubespawner/slugs.py
And this is the kind of names that get generated: 
https://github.com/jupyterhub/kubespawner/blob/main/tests/test_slugs.py#L6

PR #854, opened 2024-09-01
K8 dot unames

yhal-nesi:
Resolves the following issue https://github.com/OSC/ood_core/issues/852 by sanitizing k8 namespace name. Deals with ".", "@", unicode, and converts upper case to lower case. Slug generator is based on JupyterHub code https://github.com/jupyterhub/kubespawner/blob/main/kubespawner/slugs.py and can also be used to sanitize other types of k8 objects. 

The bootstrap script would also need to be modified to take advantage of that. https://github.com/OSC/ondemand/blob/master/hooks/k8s-bootstrap/k8s-bootstrap-ondemand.sh
Otherwise the change is not compatible. 

Most of the work done by my colleague @DininduSenanayake.


johrstrom:
I think we're overthinking this. Why not just `gsub` replace the bad characters?

> The bootstrap script would also need to be modified to take advantage of that. https://github.com/OSC/ondemand/blob/master/hooks/k8s-bootstrap/k8s-bootstrap-ondemand.sh

Another reason to simplify this. Can this script compute the same hash/various other calculations? No, but it could `tr` or `sed` search and replace.

yhal-nesi:
I guess I am just trying to make it work in all cases. Let's say we replace "." by "-", and then firstname.lastname is going to become firstname-lastname. But what if there is already a user with that username? Same with upper case to lower case conversion etc. It is a bit contrived and you could say that if someone has a directory like that it is their problem. 

> Another reason to simplify this. Can this script compute the same hash/various other calculations? No, but it could tr or sed search and replace.

The way I was testing this is to call ruby inside a script:
```
export SANITIZED_USERNAME=$(ruby -r /opt/ood-hooks/k8s-bootstrap/slug_generator.rb -e "puts(SlugGenerator.safe_slug '$ONDEMAND_USERNAME')")
export NAMESPACE="${NAMESPACE_PREFIX}${SANITIZED_USERNAME}"
```

johrstrom:
> I guess I am just trying to make it work in all cases.

I'd rather have something simple that solves most cases we actually see in the wild. I think a simple `gsub` and `tr` would do this.  Maybe we can query `Etc` for collisions? 

For example, while `€` can be a legitimate username on some systems, we'd never actually see it in practice.

I commend you on your diligence to cover all cases, but as a maintainer I need to balance simplicity with the solutions the code provides. Which is to say, I'd rather cover most cases we'll actually see with simple solutions than to have a complex one that covers a lot of cases we'll never actually encounter. (not to mention the distribution and upgrade problems this would bring)

johrstrom:
Sorry, I have to close this as being too complex. Again, it's a matter of the complexity that I have to take on and with such a large code-base and feature set I just can't accept something this complex to solve such a small problem.

That said, thanks for opening the pull request and putting the work in and I apologize that it didn't make it through. I'm just a steward of this project and as such I have to consider what future maintainers will need to deal with.

yhal-nesi:
No worries, I understand completely. I should have checked by creating an issue first. Looks like a partial fix is available in version 4 (replacing "." and "@") and that should work for us, thank you.

Issue #856, opened 2024-10-22
Slurm Metrics - sacct command to get job historic data


abujeda:
in Harvard IQSS, we are working in creating a metrics widget for Open OnDemand.
After evaluating different datasources for the metrics, we have settled for getting the metrics directly from `Slurm`.

As a proof of concept, we created an script that uses `sacct` command to extract the relevant historic data from slurm to build the required metrics for our new dashboard widget. The historic data will be for a 30 days, 7 days and last 24 periods

A summary of our metrics requirements:
 - Number of CPU/GPU jobs by State: Completed, Timeout, Canceled, Out of Memory, Failed
 - CPU Metrics: Average Used, Average Efficiency, Average Allocated, Total Walltime
 - GPU Metrics: Average Allocated, Total Usage
 - Memory Metrics: Average Used, Average Efficiency, Average Allocated, Total Used
 - Time Metrics: Average Used, Average Efficiency, Average Allocated, Average Waiting Time

As an OnDemand prototype, we have created a `Slurm` adapter extensions via a monkey patch to execute the relevant `sacct` command. As well as a widget to calculate the metrics and display the results.

We will be creating a PR for the Slurm adapter changes to support this `sacct` command as we think it would be useful for other institutions.

abujeda:
This is the OnDemand prototype that we created:
<img width="800" alt="Screenshot 2024-10-22 at 12 37 59" src="https://github.com/user-attachments/assets/e7fd8488-f48c-4f7a-b403-0de434356402">

johrstrom:
> We will be creating a PR for the Slurm adapter changes to support this sacct command as we think it would be useful for other institutions.

Historic info is something I've been thinking about too. I think it should be a new API like `historic_info` to separate the two.

abujeda:
Created a PR with our IQSS specific implementation: https://github.com/OSC/ood_core/pull/857
Struggling to make it Slurm agnostic.

PR #857, opened 2024-10-23
Added support to execute sacct to get job historic data for metrics

abujeda:
Draft implementation to add support to the Slurm adapter to execute sacct command to get historic job data to calculate metrics

Fixes: #856


abujeda:
Struggling to make it Slurm agnostic. Currently it returns the raw response from Slurm as an array of hashes.

treydock:
I don't think OnDemand supports getting job step data, so maybe need `--allocations` flag?  That won't show memory usage though, but I'd advocate utilizing things like Grafana and the current ability to link Grafana panels to a job's view as Slurm's accounting database isn't really great for collecting metrics about jobs like memory or CPU usage.  If you do not use `--allocations` maybe only care about the job ID and batch step as I believe batch step will have usage for whole job and not individual step.

abujeda:
Thanks Trey. We are only interested in overall job data, but we need the data in the job/batch steps for the memory usage.
We are currently merging all the job related data into a single row. Using the max value for CPU, Memory and Elapsed. We do the merging as part of the metrics calculations.

We could do the merging inside the adapter code and create a partially populated `Info` object. But I am not sure this merging should be done at this level.

>  I'd advocate utilizing things like Grafana and the current ability to link Grafana panels to a job's view

We will be looking at using Grafana for other metrics after completing the MVP for the Slurm metrics widget.
Could you point me to an example on how to link a Grafana panel to a job's view

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -699,6 +755,10 @@ def nodes
           @slurm.nodes
         end
 
+        def sacct_metrics(job_ids: [], states: [], from: nil, to: nil)
+          @slurm.sacct_metrics(job_ids, states, from, to)
+        end
```
johrstrom, line 760: I think we should have an actual `historic_info` API on the adapter class itself to mimic and extend the `info` API.

Not 100% sure on the method signature here, but these are all keywords so it should be OK for now.

Note that this API should probably respond with an array of Info objects (currently returns an array of hashes?).

treydock:
> We will be looking at using Grafana for other metrics after completing the MVP for the Slurm metrics widget. Could you point me to an example on how to link a Grafana panel to a job's view

https://osc.github.io/ood-documentation/latest/customizations.html?highlight=grafana#grafana-support

https://grafana.com/grafana/dashboards/12093-ondemand-clusters/

Example: 
Screenshot 2024-10-24 at 9 19 39 AM

treydock:
The `Detailed Metrics` link will open a URL to the dashboard that is using parameters to show just that job's data and nodes.

lib/ood_core/job/adapters/slurm.rb
abujeda: I made some changes to return an array of info objects and added support to disable job steps.

Still when job steps are enabled, it will return them as regular info objects. Not sure this is the best approach at this point, but for our use case, we need the steps for memory metrics calculations.

johrstrom: I think I also wanted the top level `historic_info` on the Adapter class that this is then the implementation for.

abujeda: OK - will make the changes and see how that looks.

abujeda: I updated the PR with a first implementation to add the historic info interface

johrstrom (approved):

PR #863, opened 2025-01-28
SSH Keyfile location support for systemd launcher/cluster

adamboutcher:
This should give some (albeit poor quality) support for alternative ssh keyfile locations using the systemd launcher.

Our use case is so users don't have to think about ssh key generation etc to reduce a barrier of entry.

I don't write ruby so this is probably a mess.

### DIFF
--- a/lib/ood_core/job/adapters/systemd/launcher.rb
+++ b/lib/ood_core/job/adapters/systemd/launcher.rb
@@ -131,9 +131,6 @@ class OodCore::Job::Adapters::LinuxSystemd::Launcher
 
   end
 
-
-  end
-
   def shell
     ENV['SHELL'] || '/bin/bash'
   end
--- a/lib/ood_core/job/adapters/systemd/launcher.rb
+++ b/lib/ood_core/job/adapters/systemd/launcher.rb
@@ -106,26 +106,33 @@ class OodCore::Job::Adapters::LinuxSystemd::Launcher
   # @param cmd [Array<#to_s>] the command to be executed on the destination host
   def ssh_cmd(destination_host, cmd)
 
-    basecmd=[
+  def ssh_cmd(destination_host, cmd)
+
+    sshcmd=[
       'ssh', '-t',
       '-p', OodCore::Job::Adapters::Helper.ssh_port,
       '-o', 'Batchmode=yes',
-      "#{username}@#{destination_host}",
-    ].concat(cmd)
+    ]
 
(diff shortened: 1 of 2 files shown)


lib/ood_core/job/adapters/systemd/launcher.rb
```diff
@@ -97,27 +99,34 @@ def call(cmd, *args, env: {}, stdin: "")
...
-        "#{username}@#{destination_host}"
-      ].concat(cmd)
+
+    basecmd=[
+      'ssh', '-t',
+      '-p', OodCore::Job::Adapters::Helper.ssh_port,
+      '-o', 'Batchmode=yes',
+      "#{username}@#{destination_host}",
+    ].concat(cmd)
+
```
johrstrom, line 115: I think a structure like this is more idiomatic for Ruby.

```ruby
args = ['ssh', '-t'] # plus more, just for brevity
args.concat([...]) << if strict_host_checking
args.concat([...]) << unless !ssh_keyfile.to_s.empty?

args
```

adamboutcher: I'll give it a test, I've never written ruby before so it's all new to me 😋

johrstrom: I can edit this PR, I'm happy to take care of it if you like.

adamboutcher: I think I've updated it adequately. If not please feel free to change it up.

johrstrom: Thanks, sorry for the delay. I'll get around to this soon.

adamboutcher: Yes there was, I've now removed this.

johrstrom:
Sorry for the delay - I'm still working through test cases that should have caught that issue (if there was one?)

johrstrom:
> Sorry for the delay - I'm still working through test cases that should have caught that issue (if there was one?)

🤦‍♂️  I added a test for the linux_host adapter, not the systemd adapter.

I'll get this merged shortly and add some tests for the same.

johrstrom (approved):
Thanks for working through this with me!

adamboutcher:
> Thanks for working through this with me!

No worries, I'm not well versed in git or ruby so I can only apologise for being slow and cumbersome.

PR #865, opened 2025-02-04
Handle (null) tres in str_to_queue_info.

pddro-Vestas:
If a partition has TRES=(null), the parsing in str_to_queue_info fails and auto_queues is not available. This change explicitly handles the (null) case and maps to {}, effectively avoiding the failure.


johrstrom (approved):
thanks for this fix. We'll try to get it published in 4.0.1.

pddro-Vestas:
Happy to contribute. Thank you for the amazing project!

PR #867, opened 2025-02-06
Systemd fixups

johrstrom:
Small fixup in the logic provided in #863 along with some tests for the same feature.

I probably need to rebase.


johrstrom:
I did need to rebase...

PR #885, opened 2025-07-03
Add HTCondor adapter

fodinabor:
Hi there!
We are currently using HTCondor as a scheduler on our cluster (some GPUs.. :)).

I wanted to allow our ML users to use something more interactive and thus set up OOD with VS Code server.
To support this, we obviously need an adapter, which is not present so far.
So I set out to "just quickly write one".
A big disclaimer, and I'm sorry if that makes stuff harder to review at times: I'm not a ruby guru, so I quite strongly relied on what GitHub Copilot suggested (in fact, the initial conversion was done asking Copilot to convert the Fujitsu adapter).
That being said, I did my best to test and refine the code (there was obviously a lot broken/missing.

So, now to the design:
## Job submission
I assumed the ood node does not have a schedd (is not a submission node), nor is configured to allow remote `condor_submit`.
Therefore, everything is piped to a submit host via ssh.

HTCondor does not use combined job description & run script files.
Therefore, I'm just dumping the script on the submit node `call("bash", "-c", "cat > #{path}", stdin: script)` and use the path to it as `arguments`.
The actual `executable` is a shell (bash or `script.shell_path`) that then shall execute the script.
Since the script is dumped to `/tmp/..`, I force HTCondor to transfer the script.. that shouldn't be too weird in HTCondor land.

All things that usually should be inside a submission description file are set via `-a <key>=<value>` command line arguments.
I tried covering a sensible set of parameters that apps may target (and at least cover almost all members of script that I deemed directly translatable (and immediately useful / I know how to test them).
What seems to be missing:
- `email`, `email_on_started`, `email_on_terminated`: HTCondor generally has email notification support, but I think only for termination (and errors).
- `reservation_id`: I wouldn't know of any direct mapping
- `qos`: I wouldn't know of any direct mapping

Maybe a bit unclear if this is the best mapping, is that I mapped `queue_name` to the schedd (which submission node is responsible).
I however don't really know an easy way to list all the queues (maybe `condor_q -glob` and parse the output?).

Furthermore, I am a bit sorry for ignoring the "at all costs" comment, but I'm unsure what options besides `script.native` I'd have to make things like `request_memory` configurable.

## Getting Job Information
Again via SSH, query `condor_q`, just requesting exactly the info we're interested in. We could also get JSON from it, that maybe is a more stable format to parse, but that spits out a loooot of information.

## Getting Cluster State
As with job information, just fetching what we actually need instead of all of the JSON.

## Accounts
There might not be a completely reliable way to return these.
I opted for relying on the mapping file that many people will probably use according to the [AssignAccountingGroup](https://htcondor.readthedocs.io/en/latest/admin-manual/introduction-to-configuration.html#FEATURE:ASSIGNACCOUNTINGGROUP) feature.
The file contains mappings of the form `* username group1,group2` in each line.
The path to this file (on the submission host) is configurable in the cluster config, using the key: `user_group_map`.

## Configuration
Our config looks like the following:
```yaml
job:
    adapter: "htcondor"
    bin: "/usr/bin"
    cluster: "cluster-name"
    copy_environment: false
    submit_host: "conduit2.xxx.com"
    strict_host_checking: false # hope to get that fixed when in prod :)
    additional_attributes:
      "+WantGPUHomeMounted": true
      "+WantScratchMounted": true
    default_universe: docker
    default_docker_image: ubuntu:latest
    user_group_map: /etc/condor/map_user_groups.txt
```

Most of these things are probably well known, self explanatory.
We run our cluster as a container-only cluster -> selecting the docker universe is a must.
The `additional_attributes` can be quite useful, as HTCondor allows arbitrary user-specified submission variables that can be used on the scheduler side to perform some actions. For us, this e.g. allows to say, we generally want the /home and /scratch mounted.

## Further Patches
In OSC/ondemand, I'll file the following patch, once this has gotten some feedback:
```diff
diff --git a/apps/dashboard/app/lib/account_cache.rb b/apps/dashboard/app/lib/account_cache.rb
index e27fff0ab3437d6f05fcd92666ecb05c3fdba13b..91dca682bd5efcf8d80b738f624d7a90f019bd8a 100644
--- a/apps/dashboard/app/lib/account_cache.rb
+++ b/apps/dashboard/app/lib/account_cache.rb
@@ -10,7 +10,7 @@ module AccountCache
   def accounts
     Rails.cache.fetch('account_info', expires_in: 4.hours) do
       # only Slurm support in ood_core
-      cluster = Configuration.job_clusters.select(&:slurm?).first
+      cluster = Configuration.job_clusters.find { |c| c.slurm? || c.htcondor? }
       cluster.nil? ? [] : cluster.job_adapter.accounts
     rescue StandardError => e
       Rails.logger.warn("Did not get accounts from system with error #{e}")
```

### DIFF
--- a/lib/ood_core/job/adapters/htcondor.rb
+++ b/lib/ood_core/job/adapters/htcondor.rb
@@ -21,14 +21,18 @@ module OodCore
                 bin_overrides        = c.fetch(:bin_overrides, {})
                 submit_host          = c.fetch(:submit_host, "")
                 strict_host_checking = c.fetch(:strict_host_checking, true)
-                max_walltime          = c.fetch(:max_walltime, 3600)
                 default_universe = c.fetch(:default_universe, "vanilla")
                 default_docker_image = c.fetch(:default_docker_image, "ubuntu:latest")
+                user_group_map = c.fetch(:user_group_map, nil)
+                cluster = c.fetch(:cluster, "")
                 additional_attributes = c.fetch(:additional_attributes, {})
                 htcondor = Adapters::HTCondor::Batch.new(bin: bin, bin_overrides: bin_overrides,
                     submit_host: submit_host, strict_host_checking: strict_host_checking,
-                    max_walltime: max_walltime, default_universe: default_universe,
-                    default_docker_image: default_docker_image, additional_attributes: additional_attributes,
+                    default_universe: default_universe,
+                    default_docker_image: default_docker_image, 
+                    user_group_map: user_group_map,
+                    cluster: cluster,
+                    additional_attributes: additional_attributes,
                     )
                 Adapters::HTCondor.new(htcondor: htcondor)
             end
@@ -61,10 +65,6 @@ module OodCore
                     # @return [Bool]; true if empty
                     attr_reader :strict_host_checking
 
-                    # Max walltime in seconds for jobs submitted to HTCondor
-                    # @return [Integer] the maximum walltime in seconds
-                    attr_reader :max_walltime
-
                     # Default universe for jobs submitted to HTCondor
                     # @return [String] the default universe for jobs
                     attr_reader :default_universe
@@ -72,6 +72,15 @@ module OodCore
                     # Default docker image for jobs submitted to HTCondor
                     # @return [String] the default docker image for jobs
                     attr_reader :default_docker_image
+
+                    # A path to the user/group map for HTCondor jobs
+                    # The format in the file should adhere to the format used by [AssignAccountingGroup](https://htcondor.readthedocs.io/en/latest/admin-manual/introduction-to-configuration.html#FEATURE:ASSIGNACCOUNTINGGROUP)
+                    # @return [String,nil] the path to the user/group map file
+                    attr_reader :user_group_map
+
+                    # The cluster name for this HTCondor instance
+                    # @return [String] the cluster name
+                    attr_reader :cluster
 
                     # Additional attributes to be added to the job submission
                     # @return [Hash{#to_s => #to_s}] additional attributes to be added to the job submission
@@ -84,14 +93,15 @@ module OodCore
                     # @param bin [#to_s] path to HTCondor installation binaries
                     # @param submit_host [#to_s] Submits the job on a login node via ssh
                     # @param strict_host_checking [Bool] Whether to use strict host checking when ssh to submit_host
-                    def initialize(bin: nil, bin_overrides: {}, submit_host: "", strict_host_checking: false, max_walltime: 3600, default_universe: "vanilla", default_docker_image: "ubuntu:latest", additional_attributes: {})
+                    def initialize(bin: nil, bin_overrides: {}, submit_host: "", strict_host_checking: false, default_universe: "vanilla", default_docker_image: "ubuntu:latest", user_group_map: nil, cluster: "", additional_attributes: {})
                         @bin                  = Pathname.new(bin.to_s)
                         @bin_overrides        = bin_overrides
                         @submit_host          = submit_host.to_s
                         @strict_host_checking = strict_host_checking
-                        @max_walltime         = max_walltime
                         @default_universe     = default_universe.to_s
                         @default_docker_image = default_docker_image.to_s
+                        @user_group_map       = user_group_map.to_s unless user_group_map.nil?
+                        @cluster              = cluster.to_s
                         @additional_attributes = additional_attributes
                     end
 
@@ -176,6 +186,24 @@ module OodCore
                     end
 
 
+                    # Retrieve accounts using user_group_map on @submit_host
+                    # @return [Hash{String => Array<String>}] mapping of usernames to their groups
+                    def get_accounts
+                        raise Error, "user_group_map is not defined" if user_group_map.nil? || user_group_map.empty?
+                        
+                        # Retrieve accounts and cache them
+                        output = call("cat", user_group_map)
+                        accounts = {}
+                        output.each_line do |line|
+                            next if line.strip.empty? || line.start_with?("#") # Skip empty lines and comments
+                            _, username, groups = line.strip.split(/\s+/, 3)
+                            accounts[username] = groups.split(",") if username && groups
+                        end
+                        accounts
+                    rescue Error => e
+                        raise Error, "Failed to retrieve accounts: #{e.message}"
+                    end
+
                     private
 
                     # Parse the output of `condor_q` into a list of job hashes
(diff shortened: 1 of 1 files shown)


johrstrom:
Thanks for the addition! It's going to take me a bit to look this over, and we're about to be on holiday too.

Plus we don't run htcondor nor am I familiar with it's CLI, so 

But at a glance, I'd have this feedback:
* I see you have `puts` statements here and there, I'd guess for debugging. I'd ask you remove these as we don't generally log here (even if we should, we need to figure out how to do it cleanly)
* > I assumed the ood node does not have a schedd (is not a submission node), nor is configured to allow remote condor_submit.
Therefore, everything is piped to a submit host via ssh.

  Schedulers should support both cases. Both cases being (a) when the binary is on the same machine and (b) if there's a submit_host to ssh into
* > `call("bash", "-c", "cat > #{path}", stdin: script)` 

  I'm not sure what you're doing with this, but maybe just writing a temp file with, but maybe just writing the a temp file with `Tempfile` is better than relying on pipes? (especially if you're running this over ssh where pipes can be weird when reading MOTD and similar ssh messages) 
*   Is this interface written better as an array instead of key value pairs?
```
   additional_attributes:
      "+WantGPUHomeMounted": true
      "+WantScratchMounted": true
```

like this instead? Do the values ever matter here? 
```
   additional_attributes:
      - "+WantGPUHomeMounted"
      - "+WantScratchMounted"
```

fodinabor:
Thanks for your quick feedback!
We're planning to bring this to production next week. Might have production fixes coming in then :)

> I see you have puts statements here and there, I'd guess for debugging. I'd ask you remove these as we don't generally log here (even if we should, we need to figure out how to do it cleanly)

Whoopsi, will do, tomorrow!

> Schedulers should support both cases. Both cases being (a) when the binary is on the same machine and (b) if there's a submit_host to ssh into

The wrap_ssh should still handle this as elsewhere.

> I'm not sure what you're doing with this, but maybe just writing a temp file with, but maybe just writing the a temp file with Tempfile is better than relying on pipes? (especially if you're running this over ssh where pipes can be weird when reading MOTD and similar ssh messages)

The idea is to write the script to a temp file on the submission node. Tempfile only works locally.. Thus I need a way to transfer it.. Could potentially use scp or so? Do you have a better way, that you commonly use?

> Is this interface written better as an array instead of key value pairs?

I'd keep it as is - the submission parameters are generally key value parameters: 
https://htcondor.readthedocs.io/en/latest/users-manual/submitting-a-job.html
https://htcondor.readthedocs.io/en/latest/man-pages/condor_submit.html

johrstrom:
> The idea is to write the script to a temp file on the submission node. Tempfile only works locally.. Thus I need a way to transfer it.. Could potentially use scp or so? Do you have a better way, that you commonly use?

OK that makes sense. Maybe it'll be OK for now?  I guess this all hinges on the fact that the htcondor binary doesn't allow the script to be passed in through stdin? Again, I'm not familiar with htcondor's CLI at all.

fodinabor:
Yeah exactly. `condor_submit` only takes submission "commands" as stdin (or values for them), but I dont see how we could use this to hand over the script, therefore the cat hack..

lib/ood_core/job/adapters/htcondor.rb
```diff
@@ -0,0 +1,473 @@
...
+
+                    # Retrieve accounts using user_group_map on @submit_host
+                    # @return [Hash{String => Array<String>}] mapping of usernames to their groups
+                    def get_accounts
+                        raise Error, "user_group_map is not defined" if user_group_map.nil? || user_group_map.empty?
+                        
+                        # Retrieve accounts and cache them
+                        output = call("cat", user_group_map)
+                        accounts = {}
+                        output.each_line do |line|
```
johrstrom, line 198: Seems like instead of calling `cat` you should just read the file. Casting that string to a `Pathname` you can do something like

```ruby
file = Pathname.new(user_group_map || "")
raise Error, "#{user_group_map} does not exist or cannot be read" unless file.exist? && file.readable?

file.readlines.each do |line|
  # all the other stuff, though you may need to line.chomp to get rid of the newline.
end
```

lib/ood_core/job/adapters/htcondor.rb
```diff
@@ -0,0 +1,473 @@
...
+                    # @return [String] the id of the job that was created
+                    def submit_string(args: [], script_args: [], env: {}, script: "")
+                        args = args.map(&:to_s)
+                        script_args = script_args.map(&:to_s).map { |s| s.to_s.gsub('"', "'") } # cannot do double 
+                        env = env.to_h.each_with_object({}) { |(k, v), h| h[k.to_s] = v.to_s }
+
+                        tempfile = Tempfile.new("htcondor_submit")
+                        tempfile.close
+                        path = tempfile.path
+                        tempfile.unlink # unlink the tempfile so it can be used by condor_submit
```
johrstrom, line 122: Instead of creating a Tempfile, can we just generate a unique name? I feel like file operations are expensive, wouldn't `"#{Dir.tmpdir}/#{SecureRandom.uuid}"` or something similar provide the same feature?

lib/ood_core/job/adapters/htcondor.rb
```diff
@@ -0,0 +1,473 @@
...
+                    end
+
+                    # Retrieve job information using `condor_q`
+                    # @param id [#to_s] the id of the job
+                    # @param owner [String] the owner(s) of the job
+                    # @raise [Error] if `condor_q` command exited unsuccessfully
+                    # @return [Array<Hash>] list of details for jobs
+                    def get_jobs(id: "", owner: nil)
+                        args = []
+                        args.concat ["-constraint", "ClusterId == #{id}"] unless id.to_s.empty?
```
johrstrom, line 169: The `id` here is supposed to be the job id. (at least it is in the upper layers like in the `info` API). But I see here you're using `"ClusterId == #{id}"`.

Is that a mistake or is that how htcondor uses job ids?

lib/ood_core/job/adapters/htcondor.rb
fodinabor: Hmm... I guess we could require this to be _also_ present on the OOD host...
In our setup, this file is (just) deployed to the submission nodes (aka the one that we're ssh-ing to).
Reading this via SSH, ensures that the file read here corresponds with the file actually being used on the submission nodes

lib/ood_core/job/adapters/htcondor.rb
fodinabor: `ClusterId` == top level ID of the job, where a cluster is an aggregate of jobs (i.e. array jobs)
`ProcId` would be the sub id: one can have 111002.123 as job id, where 111002 is the ClusterId and 123 the ProcId.
So far, this adapter does not deal with array jobs beyond that.

johrstrom (approved):
Thanks for this contribution! I'm going to get this merged now and added into version `0.29.0` which I'll backport to 4.0 and 3.1 shortly.

Thanks again!

PR #887, opened 2025-07-15
AllowAccounts negates DenyAccounts

simonLeary42:
https://slurm.schedmd.com/slurm.conf.html#OPT_AllowAccounts

> NOTE: If AllowAccounts is used then DenyAccounts will not be enforced.

### DIFF
--- a/lib/ood_core/job/adapters/slurm.rb
+++ b/lib/ood_core/job/adapters/slurm.rb
@@ -437,7 +437,11 @@ module OodCore
                                        hsh[:AllowAccounts].to_s.split(',')
                                      end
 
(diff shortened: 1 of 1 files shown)


simonLeary42:
It also seems curious that the default value of `:allow_accounts` is `nil` but the default value of `:deny_accounts` is `[]`

simonLeary42:
> It also seems curious that the default value of `:allow_accounts` is `nil` but the default value of `:deny_accounts` is `[]`

Actually, on second thought, that's not curious, since allowing 0 accounts is not the same as allowing all accounts.

johrstrom (approved):

PR #888, opened 2025-07-15
become aware of slurm `AllowQos` `DenyQos`

simonLeary42:
As far as I know the business of actually using `AllowQos` / `DenyQos` is out of scope for OOD core, so this PR is setting up for another PR to dashboard.


lib/ood_core/job/queue_info.rb
```diff
@@ -26,6 +28,8 @@ class OodCore::Job::QueueInfo
   def initialize(**opts)
     @name = opts.fetch(:name, 'unknown')
     @qos = opts.fetch(:qos, [])
+    @allow_qos = opts.fetch(:allow_qos, nil)
```
johrstrom: Instead of `nil` we should use an empty array `[]` so that we don't get nil errors.

johrstrom (approved):

PR #890, opened 2025-07-16
remove `queue` from `AccountInfo`

simonLeary42:
The definition for `AccountInfo.queue` is "the queue this account can use", but this is problematic:
* "this account" is already muddy waters, because one `AccountInfo` is just one slurm association, so there are other "accounts" for the same slurm account with potentially other queues (or no queue)
* as far as I know, the partition for a slurm association does not imply access, it just tells Slurm that the resource limits in that association should only be applied to jobs in that partition

I can't find any usage of this attribute. Please strike down this PR if I'm just not understanding or if this attribute is important.


lib/ood_core/job/adapters/slurm.rb
```diff
@@ -193,7 +193,8 @@ def accounts
                   name: acct,
                   qos: qos.to_s.chomp.split(','),
                   cluster: cluster,
-                  queue: queue.to_s.empty? ? nil : queue
+                  # ignore queue because it doesn't imply access, only resource limits
+                  # queue: queue.to_s.empty? ? nil : queue
```
johrstrom, line 197: There's no need for the comment here if we're going to remove it.

johrstrom:
It appears we don't use it in OnDemand, so yea maybe we can just get rid of it. And if we ever need it again, we'll just add it back.

simonLeary42:
> "this account" likely needs to represent many associations

I agree. I started work on another PR which combines the `QOS` and `queue` attributes of each `AccountInfo` into a list, but then I thought this should come first.

simonLeary42:
> > May I ask what sacctmgr -nP show users withassoc format=account,cluster,partition,qos where user=$(whoami) looks like on your system?
> 
> ```
> $ sacctmgr -nP show users withassoc format=account,cluster,partition,qos where user=$(whoami)
> pi_simonleary_umass_edu|unity||long,normal,short
> pi_tbernard_umass_edu|unity||long,normal,short
> ```

We do have associations with partitions, however, they're just special cases. For example, we have QOSes `gpu-quota-{4,8,20}` which we grant to certain accounts based on how many GPUs they paid for. Those associations are limited to the partition dedicated to the GPUs that people pooled together to buy.

simonLeary42:
> As you may surmise, I edited that file by hand to provide partition_a and partition_b because we don't use partitions like that at OSC.

Removed those edits since they would be no longer used

spec/job/adapters/slurm_spec.rb
```diff
@@ -1300,7 +1300,7 @@ def job_info(opts = {})
   describe '#accounts' do
     context 'when sacctmgr returns successfully' do
       let(:slurm) { OodCore::Job::Adapters::Slurm::Batch.new }
-      let(:expected_accounts) {["pzs0715", "pzs0714", "pzs1124", "pzs1118", "pzs1117", "pzs1010", "pde0006", "pas2051", "pas1871", "pas1754", "pas1604"]}
+      let(:expected_accounts) {["pzs1124", "pzs1118", "pzs1117", "pzs1010", "pzs0715", "pzs0714", "pde0006", "pas2051", "pas1871", "pas1754", "pas1604"]}
```
johrstrom, line 1303: Maybe we should cast to_set instead of making array comparisons?

lib/ood_core/job/adapters/slurm.rb
johrstrom: I'd say remove the comment because it points to the field `queue` that no longer exists in `AccountInfo`. In a year, it'd be a rabbit-hole to nowhere because queues no longer exist.

As to the command and column - sure we can remove them now or later.

simonLeary42:
> > You seem to have a conflicting PR in #891. Which do you want?
> 
> I would want this to be merged and then I'll resolve the conflict

They aren't mutually exclusive, after a merge conflict resolution that will still deduplicate `AccountInfo`s and combine their `.qos` attributes into one list.

johrstrom:
Hmmmm, I'm not sure. This seems unnecessary given what's in #891. It seems to me that we should do #891 if we want to keep this information and just close this? Am I missing something?

simonLeary42:
> > Hmmmm, I'm not sure. This seems unnecessary given what's in #891. It seems to me that we should do #891 if we want to keep this information and just close this? Am I missing something?
> 
> The `partition` in a Slurm association applies only to the QOS in that association. If we combine the QOS lists, then `partition` has no meaning because we don't know which QOS it applies to and which it doesn't.

Sorry, that's backwards. If an association has a partition and QOSes, then those QOSes can only be used in that partition, and OOD could actually be reading that data so that `auto_queues` / `auto_qos` / `auto_account` can prevent you from submiting an un-runnable job. This would be a bunch of extra logic, and if it were implemented I would think that it should use some other attribute besides `accounts`, because I don't think it's likely that any other job scheduler is going to have a similar structure for what an account is.

johrstrom:
I meant more like the net result after all these changes.

I'm saying, isn't the net result of `890` AND `891` just basically `891` itself? I mean like why remove the `queue` attribute here and just replace it with `queues` in 891? Why not just forgo this change altogether if we're going to turn around and change it again?

simonLeary42:
My intention was to resolve the merge conflict by deleting `queues` from 891.

simonLeary42:
At the time I wasn't sure if deleting it was a good idea, so I thought we would talk about deleting it first and then I would know what to do with 891.

johrstrom:
> At the time I wasn't sure if deleting it was a good idea, so I thought we would talk about deleting it first and then I would know what to do with 891.

It seems very two steps forward, one step back. Instead I'd prefer we just take one step forward and end up in the same place. It's easier for me to manage now, but then also easier to trace through `git blame` if changes are atomic.

simonLeary42:
ok, I merged 891 into this branch.

PR #900, opened 2025-10-02
Fall back to just hostname if hostname doesn't support -A

twhitehead:
Only the [Debian hostname program](https://manpages.debian.org/testing/hostname/hostname.1.en.html) supports the -A option. Other ones like the [net-tools](https://man7.org/linux/man-pages/man1/hostname.1.html) or [coreutils](https://man7.org/linux/man-pages/man1/hostname.1@@coreutils.html) hostname programs don't.


twhitehead:
I have added it to the `linux_host` adapter now too Thanks!

PR #901, opened 2025-10-03
rescue parse exception here and refactor just a little

johrstrom:
rescue parse exception here and refactor just a little, I'll leave comments below.

### DIFF
--- a/lib/ood_core/job/adapters/sge/batch.rb
+++ b/lib/ood_core/job/adapters/sge/batch.rb
@@ -62,16 +62,19 @@ class OodCore::Job::Adapters::Sge::Batch
   # @param owner [#to_s] the owner or owner list
   # @return [Array<OodCore::Job::Info>]
   def get_all(owner: nil)
-    listener = QstatXmlRListener.new
-    argv = ['qstat', '-r', '-xml']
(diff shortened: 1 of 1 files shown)


lib/ood_core/job/adapters/sge/batch.rb
```diff
@@ -62,16 +62,19 @@ def load_drmaa(libdrmaa_path)
...
-    listener = QstatXmlRListener.new
-    argv = ['qstat', '-r', '-xml']
-    argv.concat ['-u', owner] unless owner.nil?
-    REXML::Parsers::StreamParser.new(call(*argv), listener).parse
-
-    listener.parsed_jobs.map{
-      |job_hash| OodCore::Job::Info.new(
-        **post_process_qstat_job_hash(job_hash)
-      )
-    }
```
johrstrom, line 74: This `map` call I did refactor a little bit just to be more in line with what we typically do elsewhere (better newlines and `do end` instead of `{}` ).

```ruby
      listener.parsed_jobs.map do |job_hash| 
        OodCore::Job::Info.new(**post_process_qstat_job_hash(job_hash))
      end
```

PR #906, opened 2025-11-05
Pass Id to adapters (slurm support only)

### LINKED ISSUE #905: adapters should be passed their 'id'
We're finding in #904 and others that the adapters themselves don't have their `id`. Their `id` is currently limited to the cluster object.

We need to pass this datum to the adapter itself so that it can filter other clusters for example from `sacctmgr` calls in Slurm.

johrstrom:
Fixes #905 - at least for slurm which is the only adapter at this time that needs it.

Adds a test case and a fixture file for the same. Had to update a spec file because I added the slurm fixture file.

### DIFF
--- a/spec/cluster_spec.rb
+++ b/spec/cluster_spec.rb
@@ -132,7 +132,7 @@ describe OodCore::Cluster do
     end
 
     it "responds to slurm?" do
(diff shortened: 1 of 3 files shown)


Bubballoo3 (approved):
LGTM!

PR #911, opened 2025-12-03
Bump OodCore version to 0.31.0

Bubballoo3:


lib/ood_core/version.rb
```diff
@@ -1,4 +1,4 @@
 module OodCore
   # The current version of {OodCore}
-  VERSION = "0.30.0"
```
johrstrom: Should likely be `0.30.1` instead.

PR #914, opened 2025-12-17
Slurm: total_nodes probably shouldn't account for down/unavailable nodes

dev-sda1:
Nodes that are marked as unavailable/down in Slurm, along with their processors, shouldn't be counted in the ``total_nodes`` and ``total_processors`` figure, as this leads to misleading numbers in certain areas such as the System Status, which presents more resources available than there really are.

Have also added "down" (since down nodes are also inaccessible to the user) and changed the string match on ``gres_lines`` done in #916 to ``/ig`` to account for multiple nodes being in one of those states

### DIFF
--- a/lib/ood_core/job/adapters/slurm.rb
+++ b/lib/ood_core/job/adapters/slurm.rb
@@ -124,7 +124,7 @@ module OodCore
             node_cpu_info = call("sinfo", "-aho %F/%D/%C").strip.split('/')
             gres_length = call("sinfo", "-o %G").lines.map(&:strip).map(&:length).max + 2
             gres_lines = call("sinfo", "-ahNO ,nodehost,gres:#{gres_length},gresused:#{gres_length},statelong")
-                         .lines.uniq.reject { |line| line.match?(/maint|drain|down/ig) }.map(&:split)
+                         .lines.uniq.reject { |line| line.match?(/maint|drain|down/i) }.map(&:split)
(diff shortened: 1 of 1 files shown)


lib/ood_core/job/adapters/slurm.rb
```diff
@@ -121,16 +121,16 @@ def initialize(cluster: nil, bin: nil, conf: nil, bin_overrides: {}, submit_host
           # Get a ClusterInfo object containing information about the given cluster
           # @return [ClusterInfo] object containing cluster details
           def get_cluster_info
-            node_cpu_info = call("sinfo", "-aho %A/%D/%C").strip.split('/')
+            node_cpu_info = call("sinfo", "-aho %F/%D/%C").strip.split('/')
             gres_length = call("sinfo", "-o %G").lines.map(&:strip).map(&:length).max + 2
             gres_lines = call("sinfo", "-ahNO ,nodehost,gres:#{gres_length},gresused:#{gres_length},statelong")
-                         .lines.uniq.reject { |line| line.match?(/maint|drain/i) }.map(&:split)
```
johrstrom: Apparently `g` is not valid for ruby regular expressions.

dev-sda1: Should be sorted now, committed fix

johrstrom (approved):
Thanks! This addition will be in 4.1, coming out shortly.

PR #916, opened 2026-01-12
remove "maint" and "drain" nodes which are inaccessible to the user

kwkoepcke3:
Currently "maint" and "drain" nodes are included in the calculation for total available nodes. This can lead to there being reported nodes available when none actually are


johrstrom (approved):
Thanks!

Issue #929, opened 2026-03-23
Cluster Status app doesn't support PBSPro


mrobbert:
The PBSPro job adapter needs a cluster_info API in order to support the Cluster Status app. I have some code working in my fork of ood_core that I am working on cleaning up to provide a PR.

johrstrom:
Thanks! yea PRs welcome. 4.2 is coming real soon, so basically we'd have to get anything merged in the next two weeks or so to have it available in 4.2.

mrobbert:
Closing this issue per #930

PR #930, opened 2026-03-23
Add cluster_info API for PBSPro job adapter

mrobbert:
This PR adds the cluster_info API to the PBSPro adapter which is needed by the System Status app to display the nodes, processors, and GPUs that are active and available. It has been tested on our site's two clusters and appears to work fine.


mrobbert:
Fixes #929

johrstrom (approved):
Thanks! This will get published in 4.2.

PR #932, opened 2026-04-10
Add openstack helper

andrejcermak:
After closing https://github.com/OSC/ood_support/pull/18 i am adding it here. This code is being utilized in our OpenStack interactive applications. Tested and works fine.


johrstrom (approved):
Thanks! This will get published in 4.2.

PR #938, opened 2026-04-20
cast to string and rescue generic standard error

### LINKED ISSUE #937: TypeError not caught in `parse_time` method of slurm adapter
It looks like the following section from #928 catches the wrong error class. It seems that we expect parse_time(nil) to raise an ArgumentError, but it gets past that first layer and raises a TypeError from inside `time.rb`

https://github.com/OSC/ood_core/blob/4f58de75d235867920d6038a66b3ef7e12303cc3/lib/ood_core/job/adapters/slurm.rb#L882-L887

johrstrom:
Fixes #937 

by casting the input to a string, and rescuing the more generic StandardError. Also provides a test case for the same.

### DIFF
--- a/lib/ood_core/job/adapters/slurm.rb
+++ b/lib/ood_core/job/adapters/slurm.rb
@@ -881,8 +881,8 @@ module OodCore
 
         # safely parse date time string, return nil when there are errors.
         def parse_time(date_time)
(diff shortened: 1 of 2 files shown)


Bubballoo3 (approved):
LGTM!

PR #946, opened 2026-05-28
slurm: prevent nodehost column truncation in get_cluster_info

### LINKED ISSUE #945: System Status GPUs Available shows in-use count when node FQDNs exceed 20 chars
**Describe the bug**
On the dashboard's System Status page, the "GPUs Available" indicator
displays the number of GPUs currently in use instead of the number of
GPUs free. This affects clusters whose Slurm node names (FQDNs) are
longer than 20 characters.

Root cause is in `ood_core`. `OodCore::Job::Adapters::Slurm::Batch#get_cluster_info`
(`lib/ood_core/job/adapters/slurm.rb`) builds an `sinfo -O` format string
without specifying an explicit width for the `nodehost` field:

```ruby
gres_lines = call("sinfo", "-ahNO",
              "nodehost,gres:#{gres_length},gresused:#{gres_length},statelong")
             .lines.uniq.reject { |l| l.match?(/maint|drain|down/i) }.map(&:split)
```

sinfo -O uses a default width of 20 characters for nodehost. When
the actual hostname is longer than 20 chars, sinfo truncates it to fit and
does not insert a separating space before the next column. The
subsequent line.split then returns 3 tokens instead of 4, shifting every
column down by one:

| index | code expects | what it actually contains |
| --- | --- | --- |
| line[0] | nodehost | `<truncated-host><gres>` concatenated |
| line[1] | configured gres | **GresUsed** string |
| line[2] | gresused | **state** string (`"mixed"`, `"idle"`) |
| line[3] | state | (absent) |

Consequences:

- active_gpus = sum gpus_from_gres(line[2]) parses a state word; the regex
matches nothing evaluates to 0.
- total_gpus  = sum gpus_from_gres(line[1]) parses GresUsed evaluates
to the in-use count.

The dashboard helper then renders
GPUs Available = total_gpus - active_gpus = in_use - 0 = in_use.

**To Reproduce**
1. Configure a Slurm cluster whose NodeNames are FQDNs longer than 20 chars.
2. Submit jobs, so some GPUs are allocated and some are free.
3. Run the dashboard's System Status sinfo command directly:
sinfo -ahNO nodehost,gres:50,gresused:50,statelong
3. Observe that the first column (hostname) is truncated to 20 chars and
runs directly into the gres column with no separating whitespace.
4. Load the OnDemand dashboard's System Status page.
5. Compare "GPUs Available" with what you'd compute from the cluster's
configured GPUs minus GPUs in use.

**Expected behavior**
"GPUs Available" should equal configured_GPUs − in_use_GPUs — the number
of GPUs not currently allocated to jobs.

**Environment:**
- OS: RHEL 10
- Scheduler: Slurm
- Version: OOD 4.2.2

**Anything else?**
- Workaround: monkey-patch OodCore::Job::Adapters::Slurm::Batch#get_cluster_info
in a dashboard initializer to pass an explicit nodehost:N width to sinfo.

samoz83:

## What does this PR do?
Fixes `OodCore::Job::Adapters::Slurm::Batch#get_cluster_info` so that GPU
counts are parsed correctly on clusters with node FQDNs longer than
20 characters.

`sinfo -O` uses a default width of 20 characters for the `nodehost` field.
When a hostname exceeds that width, sinfo truncates it and emits no
separating whitespace before the next column, causing `line.split` to
return 3 tokens instead of 4. The `gres` and `gresused` columns are then
read from the wrong indices: `total_gpus` ends up summing GresUsed (the
in-use count) and `active_gpus` tries to parse the state string
(`"mixed"`, `"idle"`, …), which the GPU regex never matches, so it
evaluates to 0. The dashboard's System Status widget consequently renders
"GPUs Available" as the number of GPUs currently in use.

The fix sizes the `nodehost` column dynamically from `sinfo -ho %n`,
mirroring how `gres_length` is already calculated, so the format string
always reserves enough space to keep the columns separated regardless of
hostname length.

## Related issue
Closes #945 

## Testing
- [x] Tests included

## Checklist
- [ ] This is a large feature and was discussed in an issue first *(if applicable)*

## Anything else?
The fix has been validated on a real cluster running OOD 4.2.2 with
`ood_core` 0.31.1: applied as a dashboard initializer monkey-patch first
(to confirm), then upstreamed here. After applying, the System Status
widget shows the correct `configured − in_use` GPU count rather than the
in-use count.


johrstrom:
I'm really unsatisfied with the fact that we make multiple calls to find the max length. Can't we just set the limit to some super high number like 100? This would save us the additional `sinfo` call.  And yes we likely need to do it for the `gres_length` as well, just set it to some arbitrarily high number up front without having to make _that_ sinfo call too.

samoz83:
Pushed an update that drops both dynamic-sizing sinfo calls and uses fixed widths instead. Picked 512 for gres/gresused as it was double the longest gres string in the existing test fixture. 

Also added a regression test so future refactors don't silently re-trigger the truncation case.

johrstrom (approved):
Thanks! This looks good to me.

moffatsadeghi:
@johrstrom can this PR get merged?

johrstrom:
> @johrstrom can this PR get merged?

Yep sorry for the delay!

PR #947, opened 2026-07-09
Add PBS Pro cluster stats support

hpcvasilev:
Implement `cluster_info` for the PBS Pro adapter. `Batch#get_cluster_info` parses `pbsnodes -aSj -F json`, skips down/offline/drained/maintenance nodes, and aggregates free/total cpus and gpus into a ClusterInfo object reporting active/total processors, gpus, and nodes.

## What does this PR do?
Adds `cluster_info` support to the PBS Pro adapter, which previously fell back to the base `Adapter#cluster_info` and raised `NotImplementedError`.

`PBSPro::Batch#get_cluster_info` shells out to `pbsnodes -aSj -F json`, skips nodes reported as `down`/`offline`/`drain`/`drained`/`maint`/`maintenance`/`unknown`, and aggregates each remaining node's `ncpus f/t` and `ngpus f/t` into a `ClusterInfo` object (active/total processors, gpus, and nodes). `PBSPro#cluster_info` delegates to it — mirroring the existing `Slurm#cluster_info` / `Slurm::Batch#get_cluster_info` pattern.

This lets consumers like the OnDemand Dashboard System Status page display live CPU/GPU/node utilization for PBS Pro clusters.

## Related issue
N/A — no linked issue.

## Testing
- [x] Tests included

Adds an rspec example for `Batch#get_cluster_info` that stubs the `pbsnodes` JSON payload (a free node, a job-busy node, and a down node) and asserts the aggregated processor/gpu/node totals. `bundle exec rspec spec/job/adapters/pbspro_spec.rb` → 98 examples, 0 failures.

## Checklist
- [x] Documentation provided *(if new feature, adapter or behavior change)* — YARD doc comments on the new `get_cluster_info` / `cluster_info` methods; CHANGELOG updated
- [ ] This is a large feature and was discussed in an issue first *(if applicable)* — N/A, small self-contained adapter addition

## Anything else?
Heads-up on one line in the diff: this also adds `ostruct` as a runtime dependency in the gemspec. `ostruct` is already `require`d by `lib/ood_core/cluster.rb` but was never declared, and it's a bundled (no longer default) gem on newer Rubies. Happy to split that into a separate PR if you'd prefer to keep this one focused on the PBS Pro change.

Note this PR only produces the data. Surfacing the processor/GPU numbers on the OnDemand Dashboard System Status page needs a companion change in the [ondemand](https://github.com/OSC/ondemand) repo — adding the corresponding rows to the status view, e.g.:

```ruby
status_hash('Processors', cluster_info.active_processors, cluster_info.total_processors),
status_hash('GPUs', cluster_info.active_gpus, cluster_info.total_gpus)
```

Happy to open that follow-up PR on `ondemand` if this lands.


lib/ood_core/job/adapters/pbspro.rb
```diff
@@ -89,44 +90,43 @@ def initialize(host: nil, submit_host: "", strict_host_checking: true, pbs_exec:
...
-
-              # A node is allocated if at least one CPU has been assigned to a job
-              allocated_nodes += 1 if ncpus_assigned > 0
+            data = JSON.parse(call("pbsnodes", "-aSj", "-F", "json"))
+            nodes_h = data.fetch("nodes", {})
+
+            parse_ft = lambda do |s|
+              m = s.to_s.strip.match(/\A(\d+)\s*\/\s*(\d+)\z/)
+              m ? [m[1].to_i, m[2].to_i] : [0, 0]
             end
```
johrstrom: I think you should make this a real method instead of a lambda inside a method.

ood_core.gemspec
```diff
@@ -26,8 +26,9 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "ffi", "~> 1.16.3"
   spec.add_runtime_dependency "fog-openstack", "~> 1.1.5"
   spec.add_runtime_dependency "rexml", "~> 3.2"
+  spec.add_runtime_dependency "ostruct", "~> 0.6"
```
johrstrom: Can we remove the version requirement - I'm sure `ostruct` is stable enough.

lib/ood_core/job/adapters/pbspro.rb
```diff
@@ -89,44 +90,43 @@ def initialize(host: nil, submit_host: "", strict_host_checking: true, pbs_exec:
...
-                            total_gpus: total_gpus
+            busy_nodes = Set.new
+            available_nodes = Set.new
+
+            free_ncpus = total_ncpus = 0
+            free_gpus = total_gpus = 0
+
+            nodes_h.each do |name, info|
+              state = info.fetch("state", "").to_s.downcase
+              next if state.match?(/\b(down|offline|drain|drained|maint|maintenance|unknown)\b/)
```
johrstrom: I suspect something like this `['down', 'offline', 'drain', ...].include?(state)` may be faster than a regular expression. Claude may be able to tell you if it is.

lib/ood_core/job/adapters/pbspro.rb
```diff
@@ -89,44 +90,43 @@ def initialize(host: nil, submit_host: "", strict_host_checking: true, pbs_exec:
...
-                            total_processors: total_cpus,
-                            active_gpus: allocated_gpus,
-                            total_gpus: total_gpus
+            busy_nodes = Set.new
+            available_nodes = Set.new
+
+            free_ncpus = total_ncpus = 0
+            free_gpus = total_gpus = 0
+
+            nodes_h.each do |name, info|
```
johrstrom: I think a more rubyesque way to do this is to map nodes_h to something else, a hash containing the keys you want with the number values without having to define the variables above.

```ruby
cluster_info_data = nodes_h.values.each_with_object(Hash.new(0)) |info, data|
  state = info.fetch("state", "").to_s.downcase
  # ...
  data[:free_ncpus] += f_c
  data[:total_ncpus] += t_c
  # ...
end
```

I don't believe you need the node names, you can just 

```ruby
data[:available_nodes] += 1
data[:busy__nodes] += 1 if f_c == 0 || (f_g == 0 && t_g > 0)
```

ood_core.gemspec
hpcvasilev: Of course, removed

lib/ood_core/job/adapters/pbspro.rb
hpcvasilev: Done

lib/ood_core/job/adapters/pbspro.rb
hpcvasilev: Claude says that it will be more stable, at least. Done

lib/ood_core/job/adapters/pbspro.rb
hpcvasilev: You’re absolutely right - a map-based approach would be cleaner and more Ruby-like here. We also don’t need to retain the node names.

johrstrom:
This looks good to me, but it looks like there are merge conflicts. I can't edit it myself, so I'd guess it's the changelog or gemspec.

hpcvasilev:
It was caused by a conflict with changes in my branch from earlier experiments. I’ve resolved it now.

johrstrom (approved):
LGTM, thanks! This will be out in the next release.

PR #949, opened 2026-07-28
[HTCondor] add wallclock time, limit and server time handling

fodinabor:
## What does this PR do?
This adds attributes to the query for job state so that we get a timer and notification when the job will be out of time.

`AllowedExecuteDuration` is e.g. 3600 for 1hr jobs.
`ServerTime` just gives the current time on the HTCondor server so that we can use that as an authoritative reference.

## Testing
- [ ] Tests included
- [x] No tests included — reason: no HTCondor tests so far....

## Checklist
- [ ] ~~This is a large feature and was discussed in an issue first *(if applicable)*~~

We now get this beautiful timer :)
image


johrstrom (approved):
Thanks for the fix!

Issue #878, opened 2025-04-17
Add memory to Info object


johrstrom:
It's clear we need the [Info](https://github.com/OSC/ood_core/blob/be9ede986e129da96131ad5ab4506580198a977d/lib/ood_core/job/info.rb) objects to return memory information about the job.

Here are some thoughts I have

* units should be `bytes` so upper layers like OOD can cast and format as they like.
* We should start with total memory, I could be convinced for memory per node, but I kinda feel like we should just defer that until we need it to avoid a YAGNI situation.

Bubballoo3:
After doing some investigation, I discovered that "MinMemory"(%m) is inconsistently per node or total, depending on how the memory is requested in the first place. The following output demonstrates this fact
```
[bsingleton@ascend-login02 ~]$ sbatch --account=PZS0714 --mem=100 --cpus-per-task=3 /users/PZS0714/bsingleton/ondemand/data/sys/dashboard/projects/37nntbyn/sleep_60_min.sh
Submitted batch job 6750802
[bsingleton@ascend-login02 ~]$ squeue -j 6750802 -o %i,%m,%C
JOBID,MIN_MEMORY,CPUS
6750802,100M,3
[bsingleton@ascend-login02 ~]$ scontrol show job 6750802
JobId=6750802 JobName=sleep_60_min.sh
   UserId=bsingleton(48926) GroupId=PZS0714(5515) MCS_label=N/A
   Priority=300000538 Nice=0 Account=pzs0714 QOS=ascend-default
   JobState=RUNNING Reason=None Dependency=(null)
   Requeue=1 Restarts=0 BatchFlag=1 Reboot=0 ExitCode=0:0
   RunTime=00:02:13 TimeLimit=01:00:00 TimeMin=N/A
   SubmitTime=2026-08-06T17:20:59 EligibleTime=2026-08-06T17:20:59
   AccrueTime=2026-08-06T17:20:59
   StartTime=2026-08-06T17:21:12 EndTime=2026-08-06T18:21:12 Deadline=N/A
   PreemptEligibleTime=2026-08-06T17:21:12 PreemptTime=None
   SuspendTime=None SecsPreSuspend=0 LastSchedEval=2026-08-06T17:21:12 Scheduler=Backfill
   Partition=nextgen AllocNode:Sid=ascend-login02:584004
   ReqNodeList=(null) ExcNodeList=(null)
   NodeList=a0132
   BatchHost=a0132
[...]

Issue #870, opened 2025-02-18
Nonstandard GRES


johrstrom:
From discourse: https://discourse.openondemand.org/t/system-status-in-ood-4-0-not-showing-gpu-utilization/4024/3

Apparently `%b` may not be appropriate for GRES output. The discourse thread seems to imply `%b` is `TresPerNode` not `AllocTres` which is a more accurate accounting of the resources allocated to a job.

https://github.com/OSC/ood_core/blob/ec19744f517eb1aa7f9244d7dabf9123945dac8e/lib/ood_core/job/adapters/slurm.rb#L322

treydock:
I see no mention of `%b` in the man page for `squeue`.  If you want TRES allocated you have to use `-O/--Format` flag of `tres-alloc`.  

```
$ squeue -u $USER -O tres-alloc
TRES_ALLOC
cpu=120,mem=515456M,
```

I also see no mention of `%b` in the Slurm source code related to squeue but maybe my grep was wrong.

Another example:

```
$ squeue -t R -O tres-alloc:100
TRES_ALLOC
<SNIP>
cpu=10,mem=100G,node=1,billing=10,gres/gpu=1,gres/gpu:a100=1
```

johrstrom:
We're scheduling this to circle back and take another crack at this. Obviously refactoring everything from % formats to the longer form is going to be arduous, but even so maybe it's not as bad as I'd think? In any case, we should look into how bad it could be.

Oglopf:
Confirming this with live data. On Ascend, jobs submitted with `--gpus=N` rather than `--gres=gpu:N` leave `tres-per-node` as `N/A`:
```
JOBID     TRES_ALLOC                                        TRES_PER_NODE
7126996   cpu=21,mem=84567M,node=1,billing=21,gres/gpu=2    N/A
7126995   cpu=16,mem=64432M,node=1,billing=16,gres/gpu=1    N/A
```
Both report `gpus: 0` through `info_all`. I verified by calling the adapter directly against the live cluster.

One wrinkle for whoever fixes it: `tres-alloc` reports GPUs twice, as a typed entry and an untyped rollup (`gres/gpu:a100=16,gres/gpu=16`), so swapping in `gpus_from_gres(v[:tres_alloc])` would double count. #976 adds `gpus_from_tres` for exactly this on the sacct path; this fix probably wants to land after that.

Oglopf:
[...]

PR #954, opened 2026-08-10
Refactor to long squeue params

### LINKED ISSUE #953: Refactor squeue call from shorthand to long parameters
**What is this about?**
- [ ] Enhancement request
- [ ] OS / platform support request
- [-] Other

**Describe your request or question**
Our squeue call currently uses shorthand parameters like `%b`, mentioned in #870. However this list is a static subset of available options, as the slurm docs state
> Also see the -o, --format=<output_format> option described above (which supports greater flexibility in formatting, but does not support access to all fields because we ran out of letters)

**Why is this important to you?**
This is also necessary for #878, as I have shown in that issue that the MinMemory fetched with `%m` is ambiguously per-cpu or total memory. However fetching `tres-alloc` via named parameter (which does not have a shorthand equivalent) gives a memory value that is consistently per-node.

Bubballoo3:

## What does this PR do?
Fixes #953. Replaces the single character format flags with the longer word format flags. This opens us up to several new attributes we could include in each call, but importantly the `tres-alloc` flag, which is required for #878.

Aside from the switching of the flags, the most significant change here is to the line parsing, since there is no way to include a record separator the way we did before. Instead, we add a unit separator after every single attribute, allowing us to detect the ends of lines when the unit separator is immediately followed by a newline. Updates the fixtures to match this new expected format

## Testing
- [ ] Tests included
- [x] No tests needed — reason: This does not add new functionality, so the existing test coverage is fine

## Checklist
- [ ] This is a large feature and was discussed in an issue first *(if applicable)*

## Anything else?
For testing, you should be able to open an irb session and get identical output from `require 'ood_core'; OodCore::Job::Factory.build({adapter: 'slurm'}).info_all` and `require 'ood_core'; OodCore::Job::Factory.build({adapter: 'slurm', cluster: 'ascend'}).info_all` on this branch and master


lib/ood_core/job/adapters/slurm.rb
```diff
@@ -247,7 +247,7 @@ def squeue_required_fields
           #TODO: write some barebones test for this? like 2 options and id or no id
           def squeue_args(id: "", owner: nil, options: [])
             args  = ["--all", "--states=all", "--noconvert"]
-            args.concat ["-o", "#{RECORD_SEPARATOR}#{options.join(UNIT_SEPARATOR)}"]
+            args.concat ["-O", "#{options.push(nil).join(":#{UNIT_SEPARATOR},")}"]
```
johrstrom, line 250: Why `push(nil)` here - does it do anything? Also why change the `RECORD_SEPARATOR`?

johrstrom: I think I get the push(nil) is here - you want the `UNIT_SEPARATOR` first.

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -174,7 +174,7 @@ def get_jobs(id: "", owner: nil, attrs: nil)
               advance_past_squeue_header!(output)
 
               jobs = []
-              output.each_line(RECORD_SEPARATOR) do |line|
+              output.each_line do |line|
```
johrstrom, line 177: Why change the `RECORD_SEPARATOR`?

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -299,55 +299,55 @@ def submit_string(str, args: [], env: {})
...
+              submit_time: 'SubmitTime',
+              wckey: 'WCKey',
+              licenses: 'Licenses',
+              excluded_nodes: 'ExcNodes',
+              core_specialization: 'CoreSpec',
+              nice: 'Nice',
+              scheduled_nodes: 'SchedNodes',
+              sockets_cores_threads: 'SCT',
+              work_dir: 'WorkDir',
+              gres: 'tres-per-node',  # must come at the end to fix a bug with Slurm 18
```
johrstrom, line 350: Seems like we'd want to keep gres and make tres a new field?

johrstrom: > make tres a new field

maybe even in a separate change...

lib/ood_core/job/adapters/slurm.rb
Bubballoo3: Actually it is so that the unit separator gets included after the last element of `options`. If no length/end_char parameter is specified, each field defaults to 20 characters, which could cut off lengthy GRES values. By including the unit separator as the end character of the GRES attribute, we both override the 20 char default and provide the `"#{UNIT_SEPARATOR}\n"` sequence between lines of data

Bubballoo3: > Also why change the RECORD_SEPARATOR?

The long format flags do not accept characters before the flags, so the record_separator character was being interpreted as part of the Account flag

lib/ood_core/job/adapters/slurm.rb
Bubballoo3: The long format flags do not accept characters before the flags, so the record_separator character was being interpreted as part of the Account flag. There is no way to insert a custom character in between each line, so we just use the default of `"\n"`

lib/ood_core/job/adapters/slurm.rb
Bubballoo3: This is actually the same field, and the 'name change' is coming from slurm. Observe the following output
```
[bsingleton@ascend-login02 ~]$ squeue -j 6711855 -o %A,%b
JOBID,TRES_PER_NODE
6711855,gres/gpu:1,gres/gpu:1
```
Which is the same as the output from the new flags
```
[bsingleton@ascend-login02 ~]$ squeue -j 6711855 -O JobID,tres-per-node:25
JOBID               TRES_PER_NODE            
6711855             gres/gpu:1,gres/gpu:1    
```

johrstrom: Oooofff I see that now, even #870 says the same.

lib/ood_core/job/adapters/slurm.rb
johrstrom, line 250: I don't think we should have the `,` here - that'll definitely corrupt TRES values and likely others that have comma separated sub-strings. I could tell it's sort of wrong from the fixture file updates.

johrstrom: Also the `:` - I don't think we'd want that either.

Bubballoo3: That comma doesn't make it into the final output, due to the way these format flags are passed. Rather than special characters in a formatted template, this uses comma separated flags with options. So the Unit separator gets passed as an option, but the comma is not part of the (interpreted) separator. The only strange thing here is that there is a trailing comma here, where one might typically map the options to a string with a separator value, then join with commas, but this is simpler and there is no difference in output. Observe
```
[bsingleton@ascend-login02 ~]$ squeue -j 6711855 -O Account:@@,tres-per-node:@@,
ACCOUNT@@TRES_PER_NODE@@
pas2942@@gres/gpu:1,gres/gpu:1@@
```

Bubballoo3: This is all coming from the slurm docs on long format syntax, at https://slurm.schedmd.com/squeue.html#OPT_Format

>-O, --Format=<output_format>
Specify the information to be displayed. Also see the -o, --format=<output_format> option described above (which supports greater flexibility in formatting, but does not support access to all fields because we ran out of letters). Requests a comma separated list of job information to be displayed.
The format of each field is "type[:[.][size][suffix]]"
size
Minimum field size. If no size is specified, 20 characters will be allocated to print the information.
.
Indicates the output should be right justified and size must be specified. By default output is left justified.
suffix
Arbitrary string to append to the end of the field.

johrstrom: I see, got it. You need the comma and the `:` allows for the actual unit separator that we want.

johrstrom:
While this looks OK on the surface, I'm going to need to pull it down and have a look see on some of the pages. I'm wondering specifically about the clusterstatus page where we aggregate things and activejobs where we populate info objects (and pull stuff from the native field).

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -190,7 +190,7 @@ def get_jobs(id: "", owner: nil, attrs: nil)
                 #
                 # assuming keys and values are same length! if not we have an error!
                 line = line.encode('UTF-8', invalid: :replace, undef: :replace)
-                values = line.chomp(RECORD_SEPARATOR).strip.split(UNIT_SEPARATOR)
+                values = line.chomp("/n").split(UNIT_SEPARATOR).map(&:strip)
```
johrstrom: newlines are `\n` (forward slash) not backslash. Seems like this chomp doesn't do anything or the very last item has it (gres)?

johrstrom:
This works for me, but I'm wondering about the `\n` vs. `/n` mentioned above. I guess it's just better to catch it now even if it doesn't actually break anything.

lib/ood_core/job/adapters/slurm.rb
Bubballoo3: Yeah that definitely looks like a typo but I can look into it and see if the chomp is needed at all

Bubballoo3: I figured it out. The gres does have a "\n" at the end, and while this wasn't being chomped correctly it was being handled in the strip later on. However that strip was a remnant from dealing with extra spaces between attributes, which now is not an issue (since defining a character separator with no defined length leaves no extra space). So this has now been updated so that the chomp works correctly and the strip is removed altogether

johrstrom (approved):
Cool, looks good to me!

PR #965, opened 2026-08-21
expand queue_info

### LINKED ISSUE #955: expand queue_info

In upstream OnDemand we need lots of things from the queue information.

QueueInfo objects should return these items

* max_nodes
* min_nodes
* min_cpus (documentation should note this is per node)
* max_cpus (documentation should note this is per node)
* max_time (need to determine the unit here)

**Describe alternatives you've considered (optional)**
None

**Additional context**
None

johrstrom:

## What does this PR do?

Expands the attributes of the QueueInfo class.

## Related issue
Closes #955

## Testing
- [x] Tests included

## Checklist

## Anything else?

* I'm wondering about the `nil` defaults. Specifically around UNLIMITED and not given. I wonder if using `nil` as a default and value (to represent UNLIMITED) is a good idea.
* I'm also wondering if `max_cpus` shouldn't just be `max_cpus_per_node` directly. I think we need to consider how `auto_cores` is going to be used both in batch connect _and_ in the project manager. As in, maybe we need an `auto_cores_per_node` specifically to differentiate the use cases.
* There's also some technical debt around helper methods. For this `duration_in_seconds` is a method in the outer class Slurm, not the inner class Slurm::Batch. I just duplicated it instead of refactoring it, but it should be refactored at some point along with some other maybe class methods.

### DIFF
--- a/lib/ood_core/job/adapters/slurm.rb
+++ b/lib/ood_core/job/adapters/slurm.rb
@@ -486,6 +486,12 @@ module OodCore
                            else
                              hsh[:TRES].to_s.split(',').map { |str| str.split('=') }.to_h
                            end
+
+              hsh[:max_cpus] = hsh[:MaxCPUsPerNode]
+              hsh[:max_nodes] = hsh[:MaxNodes]
+              hsh[:min_nodes] = hsh[:MinNodes]
+              hsh[:max_time] = duration_in_seconds(hsh[:MaxTime])
+
               OodCore::Job::QueueInfo.new(**hsh)
             end
 
@@ -552,6 +558,14 @@ module OodCore
                 }.fetch(a, a)
               }.flatten
             end
+
+          # FIXME: duplicate of the outer class
+          def duration_in_seconds(time)
+            return 0 if time.nil?
+            time, days = time.split("-").reverse
+            days.to_i * 24 * 3600 +
+              time.split(':').map { |v| v.to_i }.inject(0) { |total, v| total * 60 + v }
+          end
         end
 
         # Mapping of state codes for Slurm
--- a/lib/ood_core/job/adapters/slurm.rb
+++ b/lib/ood_core/job/adapters/slurm.rb
@@ -487,9 +487,9 @@ module OodCore
                              hsh[:TRES].to_s.split(',').map { |str| str.split('=') }.to_h
                            end
 
-              hsh[:max_cpus] = hsh[:MaxCPUsPerNode]
-              hsh[:max_nodes] = hsh[:MaxNodes]
-              hsh[:min_nodes] = hsh[:MinNodes]
+              hsh[:max_cpus] = parse_max(hsh[:MaxCPUsPerNode])
+              hsh[:max_nodes] = parse_max(hsh[:MaxNodes])
+              hsh[:min_nodes] = parse_max(hsh[:MinNodes])
               hsh[:max_time] = duration_in_seconds(hsh[:MaxTime])
 
               OodCore::Job::QueueInfo.new(**hsh)
@@ -565,6 +565,12 @@ module OodCore
             time, days = time.split("-").reverse
             days.to_i * 24 * 3600 +
               time.split(':').map { |v| v.to_i }.inject(0) { |total, v| total * 60 + v }
+          end
+
+          def parse_max(max)
+            return nil if max.nil? || max.to_s == 'UNLIMITED'
+
+            max.to_i
           end
         end
 
--- a/lib/ood_core/job/queue_info.rb
+++ b/lib/ood_core/job/queue_info.rb
@@ -24,11 +24,38 @@ class OodCore::Job::QueueInfo
   # An Hash of Trackable Resources and their values.
   attr_reader :tres
 
+  # The minimum nodes a job can request for this queue.
+  # Defaults to nil.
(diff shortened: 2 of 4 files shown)


lib/ood_core/job/queue_info.rb
```diff
@@ -56,4 +83,13 @@ def gpu?
   def allow_all_qos?
     allow_qos.empty? && deny_qos.empty?
   end
+
+  private
+
+  def parse_max(max)
+    return 1 if max.nil?
+    return nil if max.to_s == 'UNLIMITED'
```
Bubballoo3: Yeah I definitely am not sure about these defaults. Why would the max default to 1 if unset? It seems like both the unset and unlimited case would have the same behavior: defer to the cluster's limitations

lib/ood_core/job/queue_info.rb
```diff
@@ -24,12 +24,39 @@ class OodCore::Job::QueueInfo
...
+  attr_reader :max_time
+
   def initialize(**opts)
     @name = opts.fetch(:name, 'unknown')
     @allow_qos = opts.fetch(:allow_qos, [])
     @deny_qos = opts.fetch(:deny_qos, [])
     @tres = opts.fetch(:tres, {})
 
+    @max_nodes = parse_max(opts.fetch(:max_nodes, 1))
+    @max_cpus = parse_max(opts.fetch(:max_cpus, 1))
```
Bubballoo3: It also seems like we are setting the 1 default in two different places. Is this necessary? Couldn't we default to nil here and `parse_max` would still convert it to 1?

lib/ood_core/job/queue_info.rb
johrstrom: OK - these have been changed to default to nil.

lib/ood_core/job/queue_info.rb
johrstrom: Yea that's not good. These have been updated to default to nil.

lib/ood_core/job/queue_info.rb
```diff
@@ -24,12 +24,40 @@ class OodCore::Job::QueueInfo
...
+  attr_reader :min_nodes
+
+  # The maximum nodes a job can request for this queue.
+  # Defaults to nil. Can return nil if unlimited.
+  # @return [Integer]
+  attr_reader :max_nodes
+
+  # The maximum CPUs the job can request. Note these are
+  # minimum per node.
+  # Defaults to nil. Can returns nil if unlimited.
```
Bubballoo3: Minor typos here and on 44 with 'Can returns'

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -572,6 +558,14 @@ def squeue_attrs_for_info_attrs(attrs)
...
               }.flatten
             end
+
+          # FIXME: duplicate of the outer class
+          def duration_in_seconds(time)
+            return 0 if time.nil?
+            time, days = time.split("-").reverse
+            days.to_i * 24 * 3600 +
+              time.split(':').map { |v| v.to_i }.inject(0) { |total, v| total * 60 + v }
+          end
```
Bubballoo3, line 568: What's the intended resolution for this duplication? Is this just a quirk that `str_to_queue_info` returns completed objects while job info returns a hash that requires an extra step to turn into an object? In my mind it almost seems like `str_to_queue_info` ought to be placed alongside `gpus_from_gres` as a class method on `Slurm`. Happy to handle that refactor separately, but curious if you are thinking along the same lines

lib/ood_core/job/queue_info.rb
```diff
@@ -56,4 +84,12 @@ def gpu?
   def allow_all_qos?
     allow_qos.empty? && deny_qos.empty?
   end
+
+  private
+
+  def parse_max(max)
+    return nil if max.nil? || max.to_s == 'UNLIMITED'
```
Bubballoo3: I don't feel super comfortable hard coding a string like 'UNLIMITED' at this level when some other scheduler could use different verbiage like 'INFINITE' or 'UNBOUNDED'. I am having trouble finding actual specs for what other schedulers might use here, but this page seems to suggest that PBS might use a lowercase 'unlimited' https://docs.adaptivecomputing.com/torque/4-0-2/Content/topics/4-serverPolicies/queueAttributes.htm#max_running. What exactly the other schedulers use is a bit of an aside, the point is that 'UNLIMITED' seems Slurm-specific and the handling of it should probably happen within the adapter

Bubballoo3 (changes requested):

lib/ood_core/job/adapters/slurm.rb
johrstrom: > What's the intended resolution for this duplication?

I was thinking a module that both the outer class and the inner class can include. But you may be suggesting pushing object creation to the upper level and having the lower layer return a hash which could work too.

lib/ood_core/job/queue_info.rb
johrstrom: OK sounds good.

lib/ood_core/job/queue_info.rb
johrstrom: Fixed to just `Returns ...`

Bubballoo3 (approved):
LGTM

PR #970, opened 2026-08-28
Add memory option to script

### LINKED ISSUE #952: Universal memory attribute on scripts
Any memory request currently has to use native attributes. Given that this is a pretty generic request available in multiple schedulers, it makes sense to have a universal attribute that individual adapters can translate, similar to `wall_time`

A single memory attribute in `Job::Script`, so that you could submit with a named argument rather than an entry in the native hash

**Additional context**
There are multiple avenues to request memory in slurm, either `mem` or `mem-per-cpu`. Careful attention will be needed to ensure that whichever one is selected here is the one with the best correlation on other schedulers

Bubballoo3:
## What does this PR do?
Adds a memory option to the script option to request memory-per-node in MiB. Adds support for this attribute in the slurm adapter

## Related issue
Fixes #952 

## Testing
- [ ] Tests included
- [x] No tests needed — reason: ___

## Checklist

## Anything else?
Still exploring documentation for other schedulers, so that will be incoming soon

### DIFF
--- a/lib/ood_core/job/adapters/psij.rb
+++ b/lib/ood_core/job/adapters/psij.rb
@@ -373,6 +373,8 @@ module OodCore
             STATE_MAP.fetch(st, :undetermined)
           end
 
+		  # Converts the MiB value from a script object to
+		  # an integer number of bytes required by ResourceSpecV1
 		  def bytes_from_memory(memory)
 			return nil if memory.nil?
 
--- a/lib/ood_core/job/adapters/psij.rb
+++ b/lib/ood_core/job/adapters/psij.rb
@@ -204,6 +204,7 @@ module OodCore
                          duration: script.wall_time,
                          custom_attributes: script.args},
             resources: {__version: 1,
+				        memory: bytes_from_memory(script.memory),
                         gpu_cores_per_process: script.gpus_per_node,
                         cpu_cores_per_process: script.cores}
           }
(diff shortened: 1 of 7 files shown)


johrstrom:
> Still exploring documentation for other schedulers, so that will be incoming soon

https://slurm.schedmd.com/rosetta.pdf

johrstrom (dismissed):
Seems fine to me.

lib/ood_core/job/adapters/psij.rb
Bubballoo3, line 1: See https://exaworks.org/psij-python/docs/v/0.9.10/.generated/index.html#resourcespecv1 as a reference for these changes. It only accepts memory as an integer number of bytes, which is why the helper method is required to take it from MiB to bytes

Bubballoo3:
@johrstrom I think I have hit all the schedulers, but let me know if I am missing any that I should add. I currently have everything on the RosettaStone doc except LoadLeveler (which doesn't have a scheduler?) plus PSIJ. Adding Fujitsu would be nice but I haven't found docs anywhere and the link from @mnakao  in https://github.com/OSC/ood_core/pull/851#issuecomment-2285347551 is no longer valid. But if you agree, I think this has enough breadth to move forward on.

johrstrom (approved):
LGTM

Issue #969, opened 2026-08-27
Empty or unset HOSTNAME in Systemd::Launcher returns empty string


Oglopf:
**Describe the bug**
The `LinuxSystemd::Launcher#parse_hostname` returns an empty string on a parse failure which then creates malformed session ids. 

**To Reproduce**
Pass an `output` with either no `HOSTNAME` at all or an empty `HOSTNAME:`.
```
launcher.send(:parse_hostname, "some output with no hostname line")
# => ""
launcher.send(:parse_hostname, "HOSTNAME:\n")
# => ""
```

**Expected behavior**
There should likely be an error raised here to let the admin know where this is happening before the ssh session tries to use the empty string and creates confusing errors far away from the root cause.

But, i'm also unsure if this was on purpose or if there is a reason for this, however `start_remote_session` interpolates the result into `"#{session_name}@#{hostname}"`, so a parse failure yields an id like `ondemand-a1b2c3d4e5@` with an empty `host`. That id is later split back apart and fed to `stop_remote_session`, which attempts an SSH to an empty hostname from what I could follow.

**Environment:**
N/A not environment specific.

**Anything else?**
This was uncovered while writing a tests for the `LinuxSystemd::Launcher`.

Issue #844, opened 2024-07-10
refactor tests to minitest


johrstrom:
While attempting to write a test case for a feature, I renewed my favor for minitest over rspec.

I just think rspec is too hard to learn and deal with when we have all sorts of other stuff going on.

So this epic ticket is to refactor all of our test suite from rspec to minitest to help promote writing test cases without having to learn rspec. Minitest has this advantage - you don't really have to learn minitest. It's just Ruby with some helpers, whereas rspec has it's own DSL that you have to learn and troubleshoot.

- [ ] bootstrap minitest

johrstrom:
@treydock I know you prefer rspec, I just think minitest is easier all around. Especially as the team grows and attempts to attract new developers, not to mention current developers' appetite to have to all these various frameworks.

I guess really - minitest just helps us narrow all the things one may have to learn/know to get anything done.

treydock:
With simplicity comes limitations, like with most things.  RSpec makes it easy to do complex stubbing (mocking) as well as support numerous kinds of matchers that read well.  Also keeping tests from spilling into other tests is a huge benefit to RSpec.

Before you fully commit to replacing RSpec, might be good to take one of two complex test cases from each RSpec file and make sure you can adequately replace them with minitest.  Simplicity at the cost of functionality is not a good trade off.  If minitest is going to make testing less reliable or function in a way that makes them less useful, that seems like a deal breaker.

Also the e2e tests in main OnDemand repo have to stay with Rspec cause there is simply no way to use minitest to replace beaker-rspec plugin.

johrstrom:
> Before you fully commit to replacing RSpec, might be good to take one of two complex test cases from each RSpec file and make sure you can adequately replace them with minitest

Yea I'm really just trying to reduce yak shaving where I can. I'll take a look at some complex test cases and see how they fit.

> Also the e2e tests in main OnDemand repo have to stay with Rspec cause there is simply no way to use minitest to replace beaker-rspec plugin.

This effort is really limited to this library, there's nothing we can do about e2e tests.

johrstrom:
>> Before you fully commit to replacing RSpec, might be good to take one of two complex test cases from each RSpec file and make sure you can adequately replace them with minitest

> Yea I'm really just trying to reduce yak shaving where I can. I'll take a look at some complex test cases and see how they fit.

[...]

PR #971, opened 2026-08-28
Add spec coverage for `LinuxSystemd::Launcher`

Oglopf:
## What does this PR do?
Adds `spec/job/adapters/systemd/launcher_spec.rb`, covering the helper methods on
`LinuxSystemd::Launcher`. This adapter currently has no test coverage.

No production code is changed — every spec asserts existing behavior.

Methods covered: `script_timeout`, `parse_hostname`, `user_script_has_shebang?`,
`script_arguments`, `export_env`, `error_path`, `unique_session_name`, `submit_host`.

## Related issue
Relates to #969  (the `parse_hostname` empty-string behavior surfaced while writing these,
but this PR does not fix it).

## Testing
- [x] Tests included

## Checklist
- [ ] Documentation provided *(if new feature, adapter or behavior change)* — N/A, no behavior change

## Anything else?
Most of these methods are private. I used `send`, following the pattern already in
`spec/job/adapters/linux_host/launcher_spec.rb`. The public surface
(`start_remote_session`, `stop_remote_session`, `list_remote_sessions`) all shells out
over SSH, so covering this branching logic through it would mean mocking `Open3` for
every case. These helpers are pure functions, so testing them directly gives clearer
failures. Happy to restructure if you'd prefer otherwise.

Two behaviors worth a look, both asserted as-is rather than changed:

- `parse_hostname` returns `''` when no `HOSTNAME:` line is present. `start_remote_session`
  interpolates that into `"#{session_name}@#{hostname}"`, producing an id ending in a bare `@`.
- `user_script_has_shebang?` checks `start_with?('#!/')`, so `#!bash` and `#! /bin/bash`
  (space after `#!`) both return false.

`bundle exec rake spec` is green locally.

### DIFF
--- a/spec/job/adapters/systemd/launcher_spec.rb
+++ b/spec/job/adapters/systemd/launcher_spec.rb
@@ -1,0 +1,230 @@
+# spec/job/adapters/systemd/launcher_spec.rb
+require "spec_helper"
+require "ood_core/job/adapters/systemd"
+require "ood_core/job/adapters/systemd/launcher"
+
+describe OodCore::Job::Adapters::LinuxSystemd::Launcher do
+  let(:opts) do 
+    {
+    :ssh_hosts => [
+      'owens-login01.hpc.osc.edu',
+      'owens-login02.hpc.osc.edu',
+    ],
+    :submit_host          => 'owens.osc.edu',
+    :site_timeout         => 20,
+    :debug                => false,
+    :strict_host_checking => true,
+    } 
+  end
+
+  # Launcher to be called in contexts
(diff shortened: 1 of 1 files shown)


Oglopf:
I have more tests for the public API that require more extensive mocking, but i'm saving that PR to push after this is done to make the review easier.

johrstrom:
Sorry but I do wonder if we want to add rspec tests. We're trying to move to minitest instead. We have #844, though haven't put any work into it, just requesting new tests be minitest.

Oglopf:
No problem at all. I'll close this out and move the tests over to `Minitest` in a separate PR.

PR #975, opened 2026-09-01
Add `Minitest` coverage for `LinuxSystemd::Launcher`

Oglopf:
## What does this PR do?
Adds tests to `test/job/adapters/systemd_launcher_test.rb`, covering the helper methods
on `LinuxSystemd::Launcher`. The file previously covered instantiation and `ssh_cmd`;
this extends it to the rest of the launcher's pure helper methods.

No production code is changed — every test asserts existing behavior.

Methods covered: `script_timeout`, `parse_hostname`, `user_script_has_shebang?`,
`script_arguments`, `export_env`, `error_path`, `unique_session_name`, `submit_host`.

This replaces #971, which used RSpec. Rewritten in Minitest per @johrstrom's comment
there, and in support of #844.

## Related issue
Relates to #969 (the `parse_hostname` empty-string behavior surfaced while writing these,
but this PR does not fix it).

## Testing
- [x] Tests included

## Checklist
- [ ] Documentation provided *(if new feature, adapter or behavior change)* — N/A, no behavior change

## Anything else?
Most of these methods are private. I used `send`, following the pattern already
established in this file's existing `ssh_cmd` tests. The public surface
(`start_remote_session`, `stop_remote_session`, `list_remote_sessions`) all shells out
over SSH, so covering this branching logic through it would mean stubbing `Open3` for
every case. These helpers are pure functions, so testing them directly gives clearer
failures. Happy to restructure if you'd prefer otherwise.

Two behaviors worth a look, both asserted as-is rather than changed:

- `parse_hostname` returns `''` when no `HOSTNAME:` line is present. `start_remote_session`
  interpolates that into `"#{session_name}@#{hostname}"`, producing an id ending in a bare `@`.
- `user_script_has_shebang?` checks `start_with?('#!/')`, so `#!bash` and `#! /bin/bash`
  (space after `#!`) both return false.

`bundle exec rake test` is green locally.

### DIFF
--- a/test/job/adapters/systemd_launcher_test.rb
+++ b/test/job/adapters/systemd_launcher_test.rb
@@ -22,8 +22,8 @@ class SystemdLauncherTest < Minitest::Test
 
   def test_ssh_cmd_default
     launcher = launcher_instance
-    expected = [  'ssh', '-t', '-p', '22', '-o', 
-                  'Batchmode=yes', '-o', 'StrictHostKeyChecking=no', '-o', 'UserKnownHostsFile=/dev/null', 
+    expected = [  'ssh', '-t', '-p', '22', '-o',
+                  'Batchmode=yes', '-o', 'StrictHostKeyChecking=no', '-o', 'UserKnownHostsFile=/dev/null',
                   'testuser@localhost', '/bin/bash'
                 ]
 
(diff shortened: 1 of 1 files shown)


johrstrom (approved):
Works for me!

PR #978, opened 2026-09-02
Gpu type info

### LINKED ISSUE #957: Add GPU type to info object
When we add the `tres-alloc` field in #878 we will get some additional information available for parsing. An example of the output from `tres-alloc` is `cpu=64,mem=257728M,node=2,billing=64,gres/gpu=4,gres/gpu:a100=4`. 

Add to GRES or new field the breakdown of which GPUs have been requested. Although most clusters will have a single type of GPU, this could be very helpful for heterogeneous clusters where there could be multiple tiers of GPU types.

Oglopf:
## What does this PR do?
Adds a `gpu_type` attribute to `Info` and populates it in the Slurm adapter's squeue
path.

`Info` currently reports `gpus` as a count with no indication of what kind. Slurm's TRES
strings carry the type when one was allocated:

```
cpu=17,mem=64G,node=1,billing=17,gres/gpu=1,gres/gpu:a100=1
                                            ^^^^^^^^^^^^^^ type and count
```

`gpu_type` is a hash of type to count — `{ 'a100' => 1 }` — rather than a single string,
so mixed-type allocations are represented honestly and the data can be extended later. It
defaults to `{}`, matching the convention for collection-valued attributes elsewhere in
`Info` and `QueueInfo#tres`.

Adds `gpu_type_from_tres` to the Slurm adapter, alongside the existing `memory_from_tres`.
Untyped entries (`gres/gpu=2`) are excluded — they carry no type information — so a job
can legitimately have a nonzero `gpus` and an empty `gpu_type`.

Scoped to Slurm. TRES is Slurm-specific, so other adapters would need their own
equivalents.

## Related issue
Closes #957

## Testing
- [x] Tests included

Six unit tests for `gpu_type_from_tres` covering typed entries, untyped-only, multiple
types, the `gres/`-less form, no GPUs, and nil/`(null)`/empty input.

One integration test through `info_all` using a new fixture,
`spec/fixtures/output/slurm/squeue_gpu_types.txt`, with three jobs: typed GPU, untyped
GPU, and no GPU. TRES formats in the fixture were taken from live squeue output on OSC
Ascend rather than invented.

`bundle exec rake test` and `bundle exec rake spec` are both green locally.

## Checklist
- [ ] Documentation provided *(if new feature, adapter or behavior change)* — attribute is
      documented inline; happy to add more if there's a preferred place

## Anything else?
Adding an attribute to `Info` required no change to `to_h`, `==`, or `eql?`, since #962
made `to_h` derive from `instance_variables`.

**Unrelated bug found while building the fixture, not fixed here:** `parse_job_info`
sources `gpus` from `tres-per-node`, which Slurm leaves as `N/A` for jobs submitted with
`--gpus=N` rather than `--gres=gpu:N`. Those jobs report `gpus: 0` despite having GPUs
allocated — I saw two such jobs running on Ascend. `tres-alloc` carries the correct count
and is already requested on this path. Filing separately.


test/job/adapters/slurm_test.rb
```diff
@@ -1,4 +1,5 @@
 require 'test_helper'
+require 'ood_core/job/adapters/slurm'
```
Bubballoo3, line 2: This doesn't seem to be necessary, is there a special case that this helps with? I can get tests to pass locally without the line, and the CI tests before this change were fine without it

Bubballoo3 (changes requested):
This looks pretty good, I only wonder if we should call this attribute `gpu_types` instead? I could see it being a potential source of confusion where someone could expect `gpu_type` to be a string, but 'types' suggests a breakdown that requires each 'type' to be extracted. Just my feeling, I could be convinced otherwise, but I want to make sure we don't end up in a situation of 'wish we'd named this better when it was first added'

Oglopf:
Totally agree on the name, great call out. That hash may have other models in there so providing a plural is semantically much clearer. I'll change it to `gpu_types` like you suggested.

test/job/adapters/slurm_test.rb
Oglopf: This took me a bit to figure out. It's load-order dependent, which is why it passes _most_ of the time. The other tests in this file reach the adapter through `Factory.build`, which requires it lazily so once any of those has run, the constant exists for everything after it in the process, but only if it came first. My new unit tests reference `OodCore::Job::Adapters::Slurm` directly, so if minitest's random order puts one of them first, you get `NameError: uninitialized constant`.

I hit it yesterday before adding the `require`, and it was very strange to hit it and took learning a bit how mintest works to debug. 2 of 6 unit tests errored and the other 4 passed in the same run, because the seed happened to work. Reproducible by running the file on its own:
```
bundle exec ruby -Ilib -Itest test/job/adapters/slurm_test.rb
```
`systemd_launcher_test.rb` requires its adapter for the same reason. Happy to drop it if you'd rather the unit tests go through `Factory.build` instead, but then it'd be an implicit dependency rather than a stated one.

Bubballoo3 (approved):
LGTM!

moffatsadeghi:
Is this PR ready to be merged?

johrstrom:
> Is this PR ready to be merged?

It appears to have a merge conflict.

Bubballoo3 (dismissed):

PR #982, opened 2026-09-10
Support nested accounts in slurm

### LINKED ISSUE #862: Slurm - account for nested accounts
The `accounts` API in slurm should account for nested accounts as they relate to partitions and their accessibly.

From discourse

https://discourse.openondemand.org/t/form-auto-queues-missing-a-partition/3958/6

Here's a partition that has a `AllowAccounts` value that's actually a parent account. Any AccountInfo that is a child of this account should be allowed in this queue.

```
PartitionName=sla-prio Nodes=basic_nodes,standard_nodes,himem_nodes TRESBillingWeights=CPU=0,Mem=0,GRES/gpu:a100=0,GRES/gpu:p100=0,GRES/gpu:v100=0,GRES/gpu:a100_3g=0,GRES/gpu:a100_1g=0,GRES/gpu=0 CpuBind=core DefaultTime=24:00:00 PriorityTier=100 ReqResv=YES AllowAccounts=reserved_allocation
```

I'm fairly sure we have to make another CLI call to accomplish this.

Bubballoo3:
## What does this PR do?
Fixes #862 by recursively querying sacctmgr to discover the hierarchy of accounts a given user falls within. This is unfortunately fairly expensive because we cannot query parent accounts that contain a given user, instead we must query by account, fetching both users and parent accounts at once (notice the blank lines in the lv0 fixture, these correspond to individual users under that account). 

This rests heavily upon the assumption that **every account is ultimately descended from root**. This is true on OSC systems as far as I can tell, but unfortunately not definitively stated in the slurm docs. Instead, we can infer it from the section on defining parent accounts via induction
> Parent=<parent>
Parent account of this account. Default is the root account, a top level account.
https://slurm.schedmd.com/sacctmgr.html#OPT_Parent

Consider a new slurm instance where `root` is the only account. Then any new account must be descending from root (whether specified or default). Then any account created next must either specify an account that has root as its parent, specify root, or use the default (root). Then any account in the next layer must specify one of those accounts, and so on. With this reasoning, I think it is a pretty absolute assumption to rest on, but if a gap in that logic exists, please point it out.

We integrate the new parent account discovery into the existing accounts method using an optional flag, allowing downstream consumers to continue using the basic account discovery unless they explicitly need parents. In OnDemand, for instance, we need all accounts for queues, but likely not in every situation.

Finally, we should consider whether this gives us all the necessary information or if we need to provide the full hierarchy in a hash object, or maybe provide both. However beyond the pure complexity of such a task, I am not even sure what format would be the most useful for something like auto_queues downstream. One possibility (from the new fixtures) would be
```
{'pas2051' => 'overhead - ohio supercomputer center',
 'pas1871' => 'overhead - ohio supercomputer center',
 'pas1754' => 'alternate - other ancestry account',
 'pas1604' => 'other: shares-parent-with-alternate',
 'overhead - ohio supercomputer center' => 'overhead',
 'alternate - other ancestry account' => 'alternate',
 'other: shares-parent-with-alternate' => 'alternate',
 'overhead' => 'root',
 'alternate' => 'root'
}
```

Hopefully these changes provide a starting place to think about what else we might need and what format we would want it to take, but since the implementation will largely depend on the target format, I wanted to wait until we got some consensus on what we want out of this.

## Testing
- [x] Tests included

## Checklist

## Anything else?
Screenshots, context, or anything reviewers should pay close attention to.


treydock:
The "top-level" account like `root` is the one with no parents, I don't think we want to assume it's named `root`.

Bubballoo3:
Great catch @treydock, it now uses the 'account with no parents' logic instead of looking for the hardcoded string `root`.

lib/ood_core/job/adapters/slurm.rb
```diff
@@ -248,6 +248,31 @@ def accounts
...
           end
 
+          # Fetches direct parents of a given set of accounts
+          def parent_accounts(accounts)
+            args = [
+                    '-nP', 'show', 'accounts', 'withassoc', 'format=parentname,qos',
+                    'where', "account='#{accounts.join("','")}'", "cluster=#{id}"
+                   ]
+            [].tap do |parents|
+              call('sacctmgr', *args).each_line do |line|
```
johrstrom, line 258: Why make 2 CLI calls here, can't we just add `parentname` to the original `sacctmgr` call?

Bubballoo3: The difference is that one is show users and one is show accounts. We also filter the initial call by username, but parent accounts are not associated with individual users, just child accounts. (Or maybe they could be, but aren't at OSC). Observe
```
[bsingleton@ascend-login02 ~]$ sacctmgr -nP show users withassoc format=account,qos,parentname where user=bsingleton cluster=ascend
pzs0714|ascend-default|
pzs1118|ascend-default|
```

The same call with `show accounts` shows all accounts on the system, since the user parameter does not effectively filter. The only way to filter a `show accounts` call is to specify a set of accounts, which we can't do without the info from the original call

Bubballoo3: The way that slurm handles (or at least displays) these parent accounts is deeply unintuitive to me though and I can't believe there isn't a CLI call to get all accounts at all levels for a given user.

treydock: This is just how Slurm handles things.  Associations don't have parents, only accounts have parents.

> I can't believe there isn't a CLI call to get all accounts at all levels for a given user.

Not for given user:

```
$ sacctmgr -nP show assoc tree format==account,parentname,user,qos
```

There is no `parentname` as the parent is shown in the tree, and Associations have no parents.  Something like the above could potentially be dumped on an interval and queried by OnDemand but that might not be great in terms of setup for sites instead of making multiple sacctmgr calls.

treydock: Also might be worth exploring the Slurm API for accounting, might have better luck dumping things and walk the JSON.

```
unset SLURM_JWT ; export $(scontrol token)

curl --get -H "Content-Type: application/json"  -H "X-SLURM-USER-NAME: $USER"  -H "X-SLURM-USER-TOKEN: $SLURM_JWT" \
https://cardinal-slurm-api.osc.edu/slurmdb/v0.0.43/user/tdockendorf -d "with_assocs=true"
```

The API docs for our version: https://slurm.schedmd.com/archive/slurm-25.05.9/rest_api.html

johrstrom: Yea I'm not against just changing the original accounts implementation from `show users` to `show accounts`. Has that been explored?


https://github.com/OSC/ood_core/blob/a2264f3734568a52c42f5344f96beda8758126c9/lib/ood_core/job/adapters/slurm.rb#L222-L225

Bubballoo3: I'll defer to Trey's better knowledge of slurm, but I don't think there is any way to filter the `sacctmgr show accounts` call by anything other than accounts. That is, if you have a query like `sacctmgr show accounts withassoc format=user,account,qos where user=bsingleton` you get every account listed, but it only tracks down the name and qos when the user matches. So I think that any 'sacctmgr show accounts' call would have to have an account filter to produce manageably sized output, and we can't get the initial set of accounts to filter against without the initial `show users` query. 

The closest thing that I can find is if we do something like `sacctmgr show associations -nP format=user,account,qos,parentname where #{CONDITION}`, where we can then swap the condition from `user=bsingleton` to `account=pzs0714,pzs1118` and reuse the same command structure for both pieces. It still wouldn't save us a call though, just reuse the template. I think my biggest reason for abandoning this path was that it felt safer to use a brand new call and not mess with the current implementation of `accounts` unless the new flag is specified. It looks like we can swap `users withassoc` for `associations` without problems, but I would hate to encounter an edge case where they have different outputs.

johrstrom: > That is, if you have a query like sacctmgr show accounts withassoc format=user,account,qos where user=bsingleton you get every account listed, but it only tracks down the name and qos when the user matches. 

I'm not sure what's wrong with that, or maybe I don't get your point. Can't we just filter the first item and keep the second in the example output below (that used `format=account,qos,parentname,user`)? If I use `time` to see how long the query takes it's generally less than 150ms.

```
pzs1117|||
pzs1118|ascend-default||bsingleton
```

Bubballoo3: I guess it just seems like we are making the parsing harder to be able to reuse a call, when the current version has two different calls that each do exactly what they need to. I am happy to take a stab at implementing your suggestion to see how it looks, but I guess I would rather let slurm handle the filtering than have to do it ourselves.

johrstrom: > I guess it just seems like we are making the parsing harder to be able to reuse a call, when the current version has two different calls that each do exactly what they need to.

Not sure how how re-usability fits here. I think the operative phrase here is _two calls_. What happens to a page load when they're both slow? It's the potential to turn a slow page load into a timeout. If 1 call is slow or creates a timeout there's nothing we can do, we have to make at least 1 call. But having the second I think opens us up to worse behavior if either one is slow.

Bubballoo3: Ok that helps me understand where you are coming from. I think you are proposing we forego any filter whatsoever and simply issue the command `sacctmgr show associations -nP format=user,account,qos,parentname`, which returns the full list of users, accounts, qos, and parents. We can then process the output from that single call to build the whole tree, rather than recursively issue new `sacctmgr` calls. In that case I think the parsing likely will still be quite complex, but I see why it is worth it to only have a single call.

johrstrom: I thought there was an issue with `show users` because the parentname of the account doesn't show up there?

What I'm suggesting is implementing this (show accounts) and filter all the empty/invalid lines.
```
sacctmgr -nP show accounts withassoc format=user,account,qos,parentname where user=$(whoami)
```

Instead of what we currently do (show users) 

https://github.com/OSC/ood_core/blob/6460ce4bb076347e82864f13dda3b6b7ad8d281d/lib/ood_core/job/adapters/slurm.rb#L224-L227

Bubballoo3: Yeah so that user filter will filter out everything but the account name (specifically will not show parent account) for accounts that don't match the filter, meaning that it cannot be used to see the associations between parent and child accounts. If we want to do it in a single call, it would have to be one without filters, which take significantly longer to execute. Here is the output using either accounts or users as the filter, and you can see that they each only provide one half of the problem. 
```
[bsingleton@ascend-login01 ~]$ sacctmgr -nP show accounts withassoc format=user,account,qos,parentname where user=bsingleton | grep pzs0714
bsingleton|pzs0714|ascend-default|
bsingleton|pzs0714|cardinal-default|
bsingleton|pzs0714|pitzer-default|
[bsingleton@ascend-login01 ~]$ sacctmgr -nP show accounts withassoc format=user,account,qos,parentname where account=pzs0714 | grep pzs0714
|pzs0714|ascend-default|overhead - ohio supercomputer center
alanc|pzs0714|ascend-default|
bsingleton|pzs0714|ascend-default|
emoffat|pzs0714|ascend-default|
johrstrom|pzs0714|ascend-default|
travert|pzs0714|ascend-default|
|pzs0714|cardinal-default|overhead - ohio supercomputer center
alanc|pzs0714|cardinal-default|
bsingleton|pzs0714|cardinal-default|
emoffat|pzs0714|cardinal-default|
johrstrom|pzs0714|cardinal-default|
travert|pzs0714|cardinal-default|
|pzs0714|pitzer-default|overhead - ohio supercomputer center
alanc|pzs0714|pitzer-default|
bsingleton|pzs0714|pitzer-default|
emoffat|pzs0714|pitzer-default|
johrstrom|pzs0714|pitzer-default|
travert|pzs0714|pitzer-default|
```
Note that the lines that show the parent association _are not associated with any user_. This means that the only way to fetch such associations is either by knowing which child account you are looking for, or by dumping the whole account tree (no filter) and looping over the output to find which ones actually matter for a specified user. Unfortunately a no filter call averages anywhere from 7-10s, which is far too long to be worth the 'single call'. If we compare that to your 150ms call with the filter (which aligns with what I have observed) there will basically never be enough levels of accounts to justify fetching them all at once.

PR #984, opened 2026-09-18
Add adapter dependency support

### LINKED ISSUE #983: Adapters should communicate whether they accept dependent jobs
Related to https://github.com/OSC/ondemand/pull/5435/, we want to be able to know in an upper layer whether a given adapter will successfully implement the `after`, `afterok`, `afternotok`, and `afterany` flags. That PR currently attempts to do that by reading the args of `submit` and comparing that with the keywords we know we are looking for. However such an approach does not work with an adapter like LHA

https://github.com/OSC/ood_core/blob/a775f2a1f76fa82ab264bef7b2db667b36d02892/lib/ood_core/job/adapters/linux_host.rb#L78-L86

which accepts the args but raises an error if any of them are set.
 It seems like rather than detecting this from an upper layer, the adapter itself should communicate whether they can actually implement these flags or not, so that an upper layer can decide to pre-emptively shield users from an error.

**Describe alternatives you've considered (optional)**
This could be something that we also use to simplify logic across adapters, as once we have a `supports_dependencies` method (or similar), we could handle the 
```
  unless (after.empty? && afterok.empty? && afternotok.empty? && afterany.empty?)
    raise JobAdapterError, 'Scheduling subsequent jobs is not available.'
  end
```
in the base adapter rather than having to implement it in each adapter separately

bstepanovski:
## What does this PR do?
Adds `supports_job_dependencies?` on job adapters (same idea as `supports_job_arrays?`) so callers can tell whether `after` / `afterok` / `afternotok` / `afterany` will actually work, instead of guessing from the `submit` method signature. That approach breaks on adapters like LinuxHost that take the kwargs but raise if you use them.

Default is `true`. Adapters that don't support deps override to `false`.

## Related issue
Closes #983 

## Testing
- [x] Tests included

## Checklist

## Anything else?
Needed for OSC/ondemand#5435 so the dashboard can hide workflows when the scheduler can't do job deps.


lib/ood_core/job/adapter.rb
```diff
@@ -133,6 +133,14 @@ def supports_job_arrays?
         true
       end
 
+      # Whether the adapter supports job dependencies via +after+, +afterok+,
+      # +afternotok+, and +afterany+
+      # @return [Boolean] - assumes true; but can be overridden by adapters that
+      #   explicitly do not
+      def supports_job_dependencies?
+        true
+      end
```
johrstrom, line 142: Most additions set this to false. Why not default to false and let most adapters use that?

johrstrom: That makes this change much smaller both in code and tests.

bstepanovski: Good point, changed it to default to false

johrstrom (approved):
LGTM

PR #985, opened 2026-09-20
Fix LinuxHost username resolution

### LINKED ISSUE #920: LinuxHostAdapter uses Etc.getlogin which returns 'root' in PUN context
### Environment
- **OOD Core Version**: 0.29.0
- **Open OnDemand Version**: 4.0.8-1
- **OS**: Rocky Linux 8
- **Ruby**: 3.3

### Problem
In `lib/ood_core/job/adapters/linux_host/launcher.rb`, the username is determined using:

```ruby
@username = Etc.getlogin
```

When running in the PUN (Per-User NGINX) context, `Etc.getlogin` returns `root` instead of the actual user. This causes the LinuxHostAdapter to **SSH as root** to the target host and **run jobs as root**.

### Impact
- **Severity**: Critical - SSH connections made as root user, they either fail or
- Jobs execute with root privileges instead of user privileges

### Root Cause
`Etc.getlogin` returns the username from the controlling terminal. In daemon/web contexts (like PUN), there is no controlling terminal, so it returns the process owner (often `root`).

From [Ruby documentation](https://www.rubydoc.info/stdlib/etc/1.8.7/Etc.getlogin):
> Unfortunately, it is often rather easy to fool `getlogin()`. Avoid `getlogin()` for security-related purposes.

### Solution
Replace with `Etc.getpwuid.name`, which correctly returns the effective user ID:

```ruby
@username = Etc.getpwuid.name
```

### Testing
**Before (incorrect):**
```ruby
# In PUN context
Etc.getlogin
# => "root"
```

**After (correct):**
```ruby
# In PUN context  
Etc.getpwuid.name
# => "username"  # actual user
```

### Note
OOD already uses `Etc.getpwuid.name` correctly elsewhere (e.g., in `session_store.rb`):
```ruby
dir = "/var/tmp/#{Etc.getpwuid.name}"
```

najibna:
## Summary

Replace `Etc.getlogin` with `Etc.getpwuid.name` when determining the
LinuxHost SSH username.

In daemon and PUN contexts, `Etc.getlogin` may return `root` instead of
the user associated with the running process. This could cause the
adapter to attempt an SSH connection using the wrong username.

## Changes

- Resolve the username using `Etc.getpwuid.name`
- Add a regression test covering differing login-session and process users

## Testing

- `bundle exec rspec spec/job/adapters/linux_host/launcher_spec.rb`

Fixes #920

Made with [Cursor](https://cursor.com)


johrstrom (approved):
LGTM, thanks!

